PR #25, opened 2018-06-29
Use app-id not id in the manifest

supertobi:
According to the documentation we should use "app-id" and not "id" in the manifest:
http://docs.flatpak.org/en/latest/manifests.html

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -1,5 +1,5 @@
 {
-    "id": "org.gimp.GIMP",
+    "app-id": "org.gimp.GIMP",
     "branch": "stable",
     "runtime": "org.gnome.Platform",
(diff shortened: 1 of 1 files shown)


Jehan:
Did this change recently? I'm pretty sure that used to be just "id".

barthalion:
Both are correct.

Jehan:
@barthalion Ok. And is one "more correct" than the other? Specifically is "id" still accepted as alias yet deprecated and changing to "app-id" is actually the way forward?

barthalion:
`app-id` is certainly more widely used. Probably both are supposed to make things easier for newcomers.

Jehan:
Ok. Let's merge this then, especially if the doc now writes down to use this one instead.

PR #26, opened 2018-07-18
GIMP 2.10.10 preparation (was: Adding SuiteSparse as dependency)

Jehan:
This is needed for alternative Matting engine: gegl:matting-levin.


Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot,build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
bot, build

Jehan:
So many builds! And I am talking to myself, means I got crazy! muhahahah!

Jehan:
bot, build

Jehan:
bot, build

Erick555:
@Jehan you may rebase this one so we will so wht's broken.

Jehan:
Ahah now all platforms fail. Apparently `org.freedesktop.Sdk.Extension.gfortran-62` extension does not exist anymore. I hope not all platform bump will be like this or this will end up extremely tiring. ;-(

Erick555:
> I hope not all platform bump will be like this or this will end up extremely tiring

They shouldn't be. This bump is different because 1.6 base (gnome 3.28 and below) and 18.08 base (gnome 3.30+) are completely different projects. That's why migrating is painful. Updating from 3.30 to 3.32 should be smooth.

Erick555:
@Jehan gfotran should be now part of Sdk, that's why extension doesn't exist. You have to drop it from manifest.

Jehan:
Actually, I just had a look at Octave manifest which also builds SuiteSparse: https://github.com/flathub/org.octave.Octave/blob/master/org.octave.Octave.yaml

I used to take some stuff from it for my previous attempt but I think I didn't take it as-is because probably back then, they were building an older version. Now that monthes passed since my initial (failed) attempt, I see they are using a more recent version. Not sure if it is the latest (haven't checked), but at this point, I am considering we should just copy this part of their manifest and not bother. If it works for them, it should for us.

We should just make sure we only build/keep only the parts of SuiteSparse needed (if I recall, this is made of many libs or something, and we don't need most of it). I mean, our flatpak is huge already and its size just increased by nearly 1/3 because of us now embedding GTK+2. :-/

Also they explain in a comment that "gfortran is part of the Freedesktop Sdk 18.08", which explains why the extension disappeared.

Erick555:
@Jehan both i386/x86_64 builds were successful but failed while uploading to flathub with random errors.

Jehan:
Are there network issues on flathub servers? I try a rebuild (apparently I cannot just request rebuild of a single platform, or even better just redoing the upload step?).

Erick555:
@Jehan build succeeded. Did you have chance to test it?

Jehan:
It does work though the flatpak is getting huger at incredible speed. I'll have a look inside to look what we can delete.

Jehan:
We are releasing GIMP 2.10.10 very soon. Let's repurpose this working branch since it's basically working, to prepare for GIMP 2.10.10 release.

Erick555:
@Jehan you may also update ghostscript to 2.27 which fixes another important security bugs.

Jehan:
> @Jehan you may also update ghostscript to 2.27 which fixes another important security bugs.

Done.

PR #33, opened 2018-10-10
manifest: Remove access to the host files

muelli:
I couldn't find a justification for accessing all the files on the host,
so I attempt to remove that access to increase the security of the sandbox.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -10,7 +10,7 @@
     "rename-icon": "gimp",
     "finish-args": ["--share=ipc", "--share=network",
                     "--socket=x11", "--socket=wayland",
-                    "--filesystem=host", "--filesystem=xdg-config/GIMP",
+                    "--filesystem=xdg-config/GIMP",
                     "--filesystem=xdg-config/gtk-3.0", "--filesystem=/tmp",
                     "--talk-name=org.gtk.vfs", "--talk-name=org.gtk.vfs.*" ],
     "tags": ["stable"],


TingPing:
It wants to be able to open all files. Should be pretty obvious.

Since its still on Gtk2 it doesn't get simple access to the file chooser portal and even then its not necessarily enough.

muelli:
fair enough. Although I believe that "all files" can be a bit more qualified.

How much opt-in should this process be? Eventually, all apps dealing with files want to open or save them. Still, they probably don't automatically get wholesale access to everything. Even if they do or did, revisiting this decision every now and then seems worthwhile.

And it's an okay thing to make the conscious decision to make certain apps have whole sale access to the file system. After all, it's not any worse than without the sandbox and the user can still opt-out by changing the permissions.

I appreciate that Gtk2 apps does not integrate nicely with the file chooser portal. And many other popular apps don't, unfortunately (like the Electron ones :-/).

For starters, though, let me propose that GIMP gets readonly access to everything and rw to $HOME rather than rw to everything. Maybe even $HOME:ro and ~/Public:rw or ~/Downloads:rw. If only to increase the incentive for people to sponsor work on the Gtk3 branch ;-) https://gitlab.gnome.org/GNOME/gimp/commits/gtk3-port

TingPing:
If you require a user to run `flatpak override --filesystem=/a/path` before ever using the app nobody will use the app and just call it garbage. A different discussion can be had in a year when portals have improved and GIMP is on a modern toolkit.

muelli:
sure. that's why I started with the proposal of home:rw and host:ro.

But yeah, I can see how the question of "How much opt-in should this process be?" can reasonably be answered with "not at all". I would add a "yet", though, to indicate that it's a matter of progress in technology, the Gtk3 port, in this case.

Feel free to close.

It'd be nice if flatpak frontends showed more explicitly that apps with wholesale access can exfiltrate your emails, cat pictures, and so one, along with instructions how to still tighten the sandbox.  But that's a discussion to be had elsewhere, e.g. on GNOME software.

TingPing:
>  that's why I started with the proposal of home:rw and host:ro.

That is a meaningless distinction. `home:rw` means you have host execution, period, all other sandboxing is meaningless for any security. `host:rw` beyond that just means you can write to like usb drives which is quite useful and adds no extra risk as mentioned.

Jehan:
I think that @TingPing pretty much summed up the situation. In other words: we'll see with GIMP 3 and portals but for now, not being able to access all files in the host would be considered a bug by most people.

Shall we close?

Jehan:
P.S.: this permission (as others) was added after people complained that they couldn't access this or that image (there are a lot of images on GNU systems, even out of home). There are indeed a lot of reasons to be able to read out of home (which is not a security issue, as @TingPing notes: you only get the permissions that the system gives you; hence if you were not allowed to write there, you still won't be allowed to write while in flatpak; flatpak does not bypass system security). People create `/data/` (or similar path) mounts for instance (e.g. which they share with other user, a common reason for this), or simply want to read an image installed by some program under `/usr/share` (which is normally meant to be a read-only directory actually purposed for any program or anyone to read its contents! So this is completely right).

muelli:
Sure, we can close this.

PR #35, opened 2018-10-21
Update poppler dependency.

Jehan:
We were using versions a bit outdated. Let's update!

Also while doing this, let's enable openjpeg as a dependency of poppler,
since this is now even a GIMP dependency as well, hence we build it
anyway. I have to set -DOpenJPEG_DIR option as it looks like poppler
build system fails to find the installed lib otherwise.


Jehan:
bot, build

PR #37, opened 2018-11-01
Update to GNOME 3.30 runtime

Jehan:


Jehan:
bot, build

Jehan:
bot, build

Jehan:
Uh! I just realize that the GNOME 3.30 Sdk doesn't have GTK+2 anymore!

@TingPing @alexlarsson Do you confirm? It's hard to check now since the manifests are not easily available anymore (see https://github.com/flatpak/flatpak/issues/2286), but a `pkg-config` run inside the Sdk at least confirms it.

Navigating a bit through flathub repos, I discover this `shared-modules` one, and in particular this:
https://github.com/flathub/shared-modules/tree/master/gtk2
Is using this as submodule the new recommended way to run a GTK+2 app in flatpak?

Erick555:
@Jehan yes, submodule is the recommended (only) way.

Jehan:
bot, build

Erick555:
There is also new ghostscript release which again fixes some security issues. You may add it here as well.
https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/tag/gs926

ramcq:
@Jehan Building this branch is destroying the Buildbot - each build is causing >1m lines of log and Buildbot stores them in RAM. It looks like warnings compiling WebKit, maybe because of new compiler in 3.30 runtime? Please test locally and don't build it on Flathub again until you work out how to disable these warnings, turn down verbosity or something! Thanks.

Erick555:
@Jehan why do you need legacy webkit here?

scx:
@Erick555 

> why do you need legacy webkit here?

This is required by the internal help browser. However, you can just disable it.
https://bugzilla.redhat.com/show_bug.cgi?id=1375809

Erick555:
@scx Thx, for the info.

@Jehan You may seriously consider dropping webkit as it only causes issues and every distro don't use it for packaging gimp.

ramcq:
At risk of being controversial, is it actually beneficial to switch the GIMP flatpak to a runtime that doesn't contain a bunch of its dependencies? Gtk2 and WebKit1 are unmaintained anyway, so is there a real difference building the old code with a new compiler vs just using an old (binary) copy from the older runtime? Or is this worthwhile because other (potentially more critical) dependencies are then maintained with security fixes etc?

Jehan:
@ramcq:
> It looks like warnings compiling WebKit, maybe because of new compiler in 3.30 runtime?

Hmmm… :-/
I have not changed anything in webkit build rules, so in any case, that's something external. Maybe the runtime update indeed.

> At risk of being controversial, is it actually beneficial to switch the GIMP flatpak to a runtime that doesn't contain a bunch of its dependencies?

Well not sure if that is controversial to someone else, but not to me. That makes sense. And to be clear, I certainly won't have the time to test or will to build webkit locally anyway (last time I tried, on my machine, it took like a **full day**!). So in any case, I am not going to update the GNOME runtime for the time being if it makes problems.

> is there a real difference building the old code with a new compiler vs just using an old (binary) copy from the older runtime?

I doubt there is any meaningful difference on GIMP code side. My reason to update the runtime was mostly maintenance, i.e. keeping dependencies up-to-date. We even had 3 duplicate dependencies (libpng, lcms2 and libwebp) which we had to build when using the runtime 3.28 because the versions in the runtime were too old for our requirements.
But well if the new runtime makes issues of its own, I guess update will have to wait.

@Erick555:
> why do you need legacy webkit here?

Because we use it? :P 
But seriously, as sad as it is regarding outdated code, we are still using GTK+2 for the time being, which also means an outdated webkit. Our master branch already uses a recent webkit, but we cannot really say when we will be able to release GIMP 3, so…
Also this is an official GIMP build, which means we are trying to keep it as vanilla and full-featured as possible. So even though @scx is right (this is an optional feature, hence we can simply not build it and GIMP will run just losing a few features), we are sticking with it while we can.

It's already a bit sad that GIMP has lost a few features in the flatpak (either because they are currently impossible inside flatpak, or because we'd need to develop dedicated code first, i.e. new portals). If even code which works were to be removed, that would be very annoying.

Erick555:
I don't think that new runtimes have compatible ABI with the old ones. We have some legacy things available as [submodules](https://github.com/flathub/shared-modules) so there is no reason to provide them as pre-compiled binaries instead.

ramcq:
> @ramcq:
> > It looks like warnings compiling WebKit, maybe because of new compiler in 3.30 runtime?
> 
> Hmmm… :-/
> I have not changed anything in webkit build rules, so in any case, that's something external. Maybe the runtime update indeed.
> 
> > At risk of being controversial, is it actually beneficial to switch the GIMP flatpak to a runtime that doesn't contain a bunch of its dependencies?
> 
> Well not sure if that is controversial to someone else, but not to me. That makes sense. And to be clear, I certainly won't have the time to test or will to build webkit locally anyway (last time I tried, on my machine, it took like a **full day**!). So in any case, I am not going to update the GNOME runtime for the time being if it makes problems.

You don't actually need to finish building WebKit to start seeing the warnings, just start it and watch the errors fly by. I expect it will be some CFLAGS tweak you can make or maybe a well-intended ./configure option which turns on a lot of -Weverything by default.

> But well if the new runtime makes issues of its own, I guess update will have to wait.

Given there are real benefits to switching for GIMP (dependencies versions etc) I do think it's worth an hour or two digging in to the configure options and the generated Makefiles that you get from WebKit, see if you can get rid of the compiler spew.

Erick555:
@Jehan security can be also an important thing for gimp users. Currently gimp is used as an example to [bash flatpak for lack of it](https://flatkill.org/).

Jehan:
@ramcq Ahahah. Well *you* hinted that it may not be that beneficial, so I was taking the lazy way out in pure bad faith! ;-)
Well I'll make a try, but not sure when. I am a developer rather than a package maintainer so my heart lies in code and the idea to play with webkit building is really not to warm my poor heart (I still remember the horrible days when I first tried to setup this manifest for webkit building; dreadful times!).

@Erick555 
> security can be also an important thing for gimp users. Currently gimp is used as an example to bash flatpak for lack of it.

We don't use webkit to load any random web page, but static page of a given domain name (gimp.org), or installed locally. So attack angles are limited. This being said, you are right that security is important. Yet if we start giving any importance to any random page on the web just because someone paid a domain name for it… we are doomed!
Most of the points in this page are moot. I didn't think I'd have to remind it on flathub bug tracker. :-)

The simple fact we use GTK+2/X makes GIMP vulnerable to various attacks (at least non-protectable). Well not more than any common rpm/deb package, but I mean we indeed have no other way than using `filesystem=host` type protection for instance (unless we want to prevent people from loading their files, which would be a funny thing to do in GIMP!).
Now we'd be happy to finish our move to GTK+3, but we cannot do anything more about this (we already do the max, which is a lot!). Well apart the fact that contributors are very welcome if anyone wants to help us finish faster!

As for security updates, well I remind a second time this is Free Software. If you spot outdated dependencies in our manifest, anyone is welcome to send a patch/make a pull request to update our flatpak.

I seriously never thought I'd have to explain this kind of things here. With all the time I spent on doing this flatpak, I don't think that GIMP is really a bad student regarding this topic.

ramcq:
> @ramcq Ahahah. Well _you_ hinted that it may not be that beneficial, so I was taking the lazy way out in pure bad faith! ;-)
> Well I'll make a try, but not sure when. I am a developer rather than a package maintainer so my heart lies in code and the idea to play with webkit building is really not to warm my poor heart (I still remember the horrible days when I first tried to setup this manifest for webkit building; dreadful times!).

Asking in GNOME hackers, @mcatanzaro suggests that adding CFLAGS=-w CXXFLAGS=-w is the quickest way to actually get the compiler to be entirely quiet during the WebKit module build. (Meanwhile @alexlarsson is working on the real problem that the Buildbot shouldn't be retaining/leaking all of these logs in RAM anyway.)

> I seriously never thought I'd have to explain this kind of things here. With all the time I spent on doing this flatpak, I don't think that GIMP is really a bad student regarding this topic.

Thank you for your work on this Jehan, as a GIMP user and Flathub contributor, I appreciate both having the fresh releases available as a Flatpak, and having such a great app as GIMP on Flathub. :)

Jehan:
> Asking in GNOME hackers, @mcatanzaro suggests that adding CFLAGS=-w CXXFLAGS=-w is the quickest way to actually get the compiler to be entirely quiet during the WebKit module build.

Thanks. I'll try to make some tests at some point for this.

> Meanwhile @alexlarsson is working on the real problem that the Buildbot shouldn't be retaining/leaking all of these logs in RAM anyway.

Cool!

> Thank you for your work on this Jehan, as a GIMP user and Flathub contributor, I appreciate both having the fresh releases available as a Flatpak, and having such a great app as GIMP on Flathub. :)

Thanks for the kind words! :-)

PR #50, opened 2019-03-01
Update to gnome 3.30 runtime

palvarez89:


palvarez89:
Sorry, already being handled here: https://github.com/flathub/org.gimp.GIMP/pull/37

Closing because it doesn't look straightforward.

Jehan:
Yeah the title was not straightforward on #37. It used to not be about GNOME runtime.

Anyway I updated the title of the other PR.

PR #54, opened 2019-04-09
Update to runtime 3.32.

Jehan:
Hopefully just bumping the runtime version is enough.


Erick555:
Smooth as silk 😄

Erick555:
@Jehan is anything blocking this one?

Jehan:
@Erick555 Not really. I was just wondering if there were no more commits I wanted to add first. Problem is that every merge results in an update of the released binaries, which means dozens of thousands of downloads for basically no perceivable changes (here the runtime update changes nothing for people, no bug fix, no features, nothing). It just feels counter-productive.

So I prefer to bundle together various changes so that at least, the update is not completely meaningless. Therefore I'll sit a bit on this one for now. It will go in, of course, no problems. There is just no hurry. :-)

Erick555:
That's reasonable however with ostree deduplication the download size of update with no changes should be pretty minimal.

palvarez89:
Upgrading it would mean that I can uninstall `org.gnome.Platform/x86_64/3.30` :)

Jehan:
Ok let's merge this one! I added some patches to our GTK+ so that this update at least makes some improvements from the vanilla GTK+2 (since we have a few patches of our own which get never included; doesn't look like GTK+2 is much maintained anymore these days).

palvarez89:
Just as a side note: This update took around 5 seconds on my laptop :)

Thanks for updating it!

Issue #22, opened 2018-06-04
Where should 3rd party plugins install?


bootchk:
I just installed 2.10.2 from flatpak onto Ubuntu 17.10.  Then I built resynthesizer (a set of 3rd party plugins, some compiled C, some .py) from its origin, which uses automake toolchain a la the ancient gimp-plugin-template. It installed, but GIMP didn't find it.  Even after I added the path to the installed locations to Edit>Preferences>Folder>Plugins and restarted.  GIMP only found it when I installed to the ~/.vars/..../plug-ins (the sandbox?).  Also, I don't understand the second location in the default plugin paths: "/app/lib/gimp/2.0/plug-ins" since my machine doesn't have an /app dir.  I expected the old path "~/gimp2.<x>/plug-ins" in the list of default plugin paths.  I could easily have made a mistake though.

What is the recommended location for installing 3rd party plugins?  If installed to the first default location, will they be trashed by an update to GIMP?  If should be installed to the second default location, where is it (what is the absolute path?)  If should be installed to a third location, will GIMP ever default to searching it, or will the user always need to add plugin paths preferences?

(I did see the other issue about flatpaks for 3rd party plugins.  I am considering building such a flatpak for resynthesizer, it that is necessary, or wait til another 3rd party plugin such as G'mic does similar.)

(I am new to flatpaks and XDG but have some experience with sandboxes on OSX)

supertobi:
GIMP should add an extension point for plugins and then you could build the plugin as a
flatpak extension. I think every thing else will lead to problems.
Have a look into this blog post:
https://blog.tingping.se/2018/03/18/flatpaking-plugins.html

Jehan:
Please feel free to make a merge request for our manifest allowing such feature. If you could also provide an example GIMP plug-in as flatpak extension as a separate manifest for me to test if this works, it would waste less time.

supertobi:
I have here a example script an the extended org.gimp.GIMP.json. In the org.gimp.GIMP.json I've added the section "add-extensions" and in the modules part I've added a directory where I install the scripts:
```
    "add-extensions": {
        "org.gimp.GIMP.Script": {
        "version": "2",
        "directory": "extensions",
        "merge-dirs": "merge",
        "subdirectories": true,
        "no-autodownload": true,
        "autodelete": true
        }
...
      {
            "name": "extensionScripts",
            "buildsystem": "simple",
            "build-commands": [
                "install -d /app/extensions/."
            ]
        }
[...]

PR #55, opened 2019-04-14
bundle gmic

eszlari:


Jehan:
I am sorry, this won't happen. I have nothing against G'MIC (quite the opposite!) and even work with its developers. But this is the official GIMP flatpak. It is meant to contain GIMP vanilla (as feature-full as possible, and only GIMP core and core plug-ins). We cannot support adding third-party plug-ins here.

Now what can be done is adding an extension point in GIMP (see #22), then you could create a separate flatpak manifest for G'MIC, to be installed through this extension point. You'd be free to maintain this other flatpak yourself (or maybe propose the maintenance to G'MIC developers, though I am not sure at all they have the additional time and would probably appreciate your contribution for maintenance).
So yeah hopefully we will get the extension point soon. That's the current blocker.

In the meantime, sorry but I will close this pull request because I just can't accept the addition. :-/

eszlari:
I have bundled gmic with Krita's flatpak and I can't say, that it increased the maintenance effort (it's only one module more). I know that it is not practical to bundle every plugin out there, but why not the major ones, that most people are using? That's not different from how other flatpaks handle it.

Jehan:
> I have bundled gmic with Krita's flatpak and I can't say, that it increased the maintenance effort (it's only one module more)

Well it's always easier to say when you are not the one who is going to maintain. ;-)

Anyway I don't know about Krita policy and this is not our problem. Our policy is to stick to provide vanilla packages (you'll notice our Windows installer or macOS DMG don't provide G'MIC either). This is the main reason and this already is enough to refuse this request, sorry.

> I know that it is not practical to bundle every plugin out there, but why not the major ones, that most people are using?

And who decides what is major or not? I can tell you for instance that we use GIMP daily professionally here and we don't have the same needs. :-)
Also I don't know if you realize that our flatpak is already quite huge. The more it goes, the bigger it is. Now as the GNOME runtime doesn't even package GTK+2 in particular, this is one more big dependency we have to bundle! So adding now even third-party plug-ins (not used by everyone) is even less a good idea (not that it would have been anyway even without this situation; as I said: vanilla).

> That's not different from how other flatpaks handle it.

Once again, not our problem. We are not here to copy what others are doing.

Anyway I won't budge. This just won't be accepted. Now be assured that I have been seriously looking at this issue in general, in particular these last few months. I am working on better plug-in support in GIMP as a general matter. This includes thinking on how to get **all** plug-ins to be more easily installed and work on **every** platform.
Basically I want generic solutions. Just making exception for every plug-in which someone find nice is not a good long term solution. What about the next time someone comes here with some pull request for some other plug-in, telling me how it is a major one? And then the next-next time with yet another plug-in. These are not theoretical **if**. We have these requests regularly in the main tracker itself to embed more plug-ins into core GIMP, and I can tell everyone has one's own view of what is a "major" plug-in. It's far from an universal concept.

eszlari:
> And who decides what is major or not?

As you know from reddit and twitter, user already said, that missing gmic is a reason for them not to use the flatpak:

https://www.reddit.com/r/GIMP/comments/bbke0h/what_are_the_limitations_of_flatpak_gimp/
https://twitter.com/PedroPS/status/1116079097317351425

Maybe just open a poll on twitter and ask them directly, what plugins they use?

> Also I don't know if you realize that our flatpak is already quite huge.

```
$ flatpak remote-info flathub org.libreoffice.LibreOffice
Download size: 265.8 MB
Installed size: 679.6 MB

$ flatpak remote-info flathub org.freecadweb.FreeCAD
Download size: 248.1 MB
Installed size: 744.1 MB

$ flatpak remote-info flathub org.kde.krita
Download size: 112.7 MB
Installed size: 253.6 MB

$ flatpak remote-info flathub org.gimp.GIMP
Download size: 108.6 MB
Installed size: 307.6 MB
```

> Now as the GNOME runtime doesn't even package GTK+2 in particular, this is one more big dependency we have to bundle!

...which will be solved, when you have ported GIMP to GTK3.

Jehan:
I gave the solution in my first comment for how you can cleanly package G'MIC for GIMP on flatpak.

hfiguiere:
Working on it.

PR #61, opened 2019-10-23
Update to GNOME 3.34 runtime.

Jehan:


Erick555:
@Jehan you need to update submodules and add intltool from there. Also rename python2 submodule, and bump ibus-gtk to match version from submodules.

Jehan:
@Erick555 Thanks. Yeah I assumed the first try would fail and I would have to update the changes. ;-)

I did all the proposed changes except for the ibus-gtk one as I did not understand what you meant I had to match with the shared modules.

Erick555:
@Jehan 

Currently you use [ibus-gtk2 1.5.19](https://github.com/flathub/org.gimp.GIMP/blob/a752f7dfb592c8955af456e57d8dc5712d921646/org.gimp.GIMP.json#L140)
gtk2 from shared-modules has [ibus-gtk2 1.5.21](https://github.com/flathub/shared-modules/blob/eb9db91c3fc5fd2e4ab4f0edbf9fcefa239c8ca2/gtk2/gtk2.json#L103).

gnome 3.34 runtime has [ibus-gtk3 1.5.20](https://gitlab.gnome.org/GNOME/gnome-build-meta/blob/gnome-3-34/elements/sdk/ibus.bst#L4). Historically mismatches between ibus-gtk2/gtk3 versions caused trouble so I recommend to update ibus-gtk2 to 1.5.20 which will match runtime. Previously I thought gnome runtime had the same versions as shared-modules but it seems not. Only freedesktop-sdk 19.08 has 1.5.21 so it's quite a mess.

BTW: pygtk seems to fail to build with new pango from runtime:
```
pango.c:8650:55: error: ‘PANGO_TYPE_FONTSET_SIMPLE’ undeclared (first use in this function); did you mean ‘PANGO_TYPE_FONT_FAMILY’?
 8650 |     pygobject_register_class(d, "PangoFontsetSimple", PANGO_TYPE_FONTSET_SIMPLE, &PyPangoFontsetSimple_Type, Py_BuildValue("(O)", &PyPangoFontset_Type));
      |                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                       PANGO_TYPE_FONT_FAMILY
```

Jehan:
> BTW: pygtk seems to fail to build with new pango from runtime:

Argh so annoying. So it would seem that this is now deprecated: https://gitlab.gnome.org/GNOME/pango/commit/d21ea13a22b1cab8c9cb604aa3fcd2ca8073befd

It means we'd have to build a custom pango with `-UPANGO_DISABLE_DEPRECATED`. So so very annoying. I am tired already of this update. I think I will let it on the side for now and come back to it in a few weeks (or more). Please anyone, if you wish to continue from where I left it, feel free to propose patches/pull requests. :-)

Jehan:
Obsoletes by #64.

PR #65, opened 2020-02-08
Sync a bit with shared-modules/gtk2/gtk2.json.

Jehan:
When building ibus-gtk2, in somme shell commands, it uses
${FLATPAK_DEST}. I don't think it is big of a problem to use old style
absolute directory (probably stays the same), but it's a good practice
and it's nice we have this variable in our manifest to even just know it
exists for other parts of the build.
Also it's better to keep the gtk2 shared module which is (mostly) copied
into our own manifest as close as possible.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -133,8 +133,8 @@
                         "/lib/gtk-2.0/2.10.0/immodules.cache"
                     ],
                     "post-install": [
-                        "install -m644 --target-directory=/app/lib/gtk-2.0/2.10.0/immodules client/gtk2/.libs/im-ibus.so",
-                        "gtk-query-immodules-2.0 > /app/lib/gtk-2.0/2.10.0/immodules.cache"
+                        "install -m644 --target-directory=${FLATPAK_DEST}/lib/gtk-2.0/2.10.0/immodules client/gtk2/.libs/im-ibus.so",
+                        "gtk-query-immodules-2.0 > ${FLATPAK_DEST}/lib/gtk-2.0/2.10.0/immodules.cache"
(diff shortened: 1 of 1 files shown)


TingPing:
Curious why your gtk2 patches aren't upstreamed? Are they generic enough to benefit others?

Jehan:
@TingPing These are definitely generic enough and could be upstream.

We kind of abandoned the hope of having anything merged within GTK+2 which is unmaintained as far as we can see (officially I have not read so, but factually none seems to care about any bug we report anymore or about our patches — when they are GTK+ **2** only that is — so they never get merged).

Anyway in the ones from our flatpak manifest, one has been [contributed nearly 2 years ago as patch files](https://gitlab.gnome.org/GNOME/gtk/issues/1052), then again more than a year ago, [this time as a merge request](https://gitlab.gnome.org/GNOME/gtk/merge_requests/518) (in case the contribution format was the problem).

The other patches, I believe they were not even contributed (cf. the [bug report on GIMP side](https://gitlab.gnome.org/GNOME/gimp/issues/2828)) though I think it's mostly because our experience (with many bug reports on GTK+2) makes that we expect we will just waste our time writing nice bug reports or comments to not have any answer. (note: no complaining here, I also understand that GTK+2 is old and uninteresting for most GTK+ devs; just stating our experience).

We also have a few GTK+2 patches for Windows and macos which are only applied in our binary packages, if I recall.

TingPing:
I'm not against the shared-module growing a few patches.

Jehan:
Right. I could make a merge request for the shared-module then. I'd be happy our flatpak to slim down a bit. :-)

PR #70, opened 2020-03-31
Update dependencies and housekeeping

corecontingency:
Update ibus-gtk2, which fixes CVE-2019-14822. Also update poppler,
ghostscript, libmypaint, and openBLAS.

Add a .gitignore, fix some incorrect tags, and update several downloads to
use https.


Jehan:
Awesome, thanks!

PR #74, opened 2020-06-02
GIMP beta branch

Jehan:
We only need the interpreter binary for running Lua plug-ins, not the
library.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -291,6 +291,19 @@
                             "sha256": "eac6b56fcda95b0fe0123849c96c8759d832ec9baded2c9c0a5b5faeffb59005"
                         }
                     ]
+                },
+                {
+                    "name": "libx265",
+                    "buildsystem": "cmake",
+                    "subdir": "source",
+                    "sources": [
+                        {
+                            "type": "git",
+                            "url": "https://github.com/videolan/x265.git",
+                            "tag": "3.4",
+                            "commit": "07295ba7ab551bb9c1580fdaee3200f1b45711b7"
+                        }
+                    ]
                 }
             ],
             "sources" : [
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -4,15 +4,6 @@
     "runtime": "org.gnome.Platform",
     "runtime-version": "3.36",
     "sdk": "org.gnome.Sdk",
-    "sdk-extensions": [
-        "org.freedesktop.Sdk.Extension.toolchain-arm"
-    ],
-    "add-extensions": {
-        "org.freedesktop.Platform.Compat.arm": {
-            "directory": "lib/arm-unknown-linux-gnueabihf",
-            "version": "19.08"
-        }
-    },
     "command": "gimp-2.99",
     "separate-locales": false,
     "rename-desktop-file": "gimp.desktop",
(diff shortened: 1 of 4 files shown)


Jehan (commented):
Argh this 2-step review procedure is really annoying. I just discover that these questions I posted in June were actually… never posted! It's mostly outdated as I moved to the beta version of luajit which doesn't have these toolchain extension requirements. But still posting these, to at least get some information on how to use sdk-extensions the next time it occurs.

Jehan:
Review for myself (because I'm not going to fix these today):

* Lua is not working, I get such errors:

```
luajit: ...mp/2.99/plug-ins/goat-exercise-lua/goat-exercise-lua.lua:22: module 'lgi' not found:
	no field package.preload['lgi']
	no file './lgi.lua'
	no file '/app/share/luajit-2.1.0-beta3/lgi.lua'
	no file '/usr/local/share/lua/5.1/lgi.lua'
	no file '/usr/local/share/lua/5.1/lgi/init.lua'
	no file '/app/share/lua/5.1/lgi.lua'
	no file '/app/share/lua/5.1/lgi/init.lua'
	no file './lgi.so'
	no file '/usr/local/lib/lua/5.1/lgi.so'
	no file '/app/lib/lua/5.1/lgi.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
	[C]: in function 'require'
	...mp/2.99/plug-ins/goat-exercise-lua/goat-exercise-lua.lua:22: in main chunk
	[C]: at 0x55afbc93c230
GIMP-WARNING: gimp-2.99: gimp_wire_read(): error
```

* Localization not working either, and I get such errors:

```
(gimp-2.99:2): Gtk-WARNING **: 21:53:31.366: Locale not supported by C library.
	Using the fallback 'C' locale.
```

Note: I saw similar errors on the stable build too. I wonder if localization is working on our main flatpak. I will have to check.

Jehan:
Ok, all but localization seems to work now.

TheEvilSkeleton:
I'd like to propose a change in the filesystem permissions:

```diff
- "--filesystem=host", "--filesystem=xdg-config/GIMP",
+ "--filesystem=xdg-pictures",
+ "--filesystem=xdg-config/GIMP",
```

Jehan:
> I don't see the point in giving --filesystem=host in an image manipulation software. Just giving it access to pictures (and maybe few other permissions) should work.

This is a software which heavily works with file. We are using it daily and professionally on various files in complex project directory structures. Do you save the source repositories of software you develop (if you are a developer) in `XDG_DOCUMENTS_DIR`? I certainly don't. Same as we don't save our huge project repositories inside `XDG_PICTURES_DIR` neither.
GIMP is not a software made solely to edit screenshots or fun one-shot images. It is massively used for advanced purpose (and not only by us).

So no we can't and won't remove `--filesystem=host`. This will only happen when we will be able to use the file portal, which we want to do eventually, but will likely not happen immediately (probably not even for GIMP 3.0) because it has limitations which would make us lose features (we need to redesign at the very least the export process to get there, and maybe other features as well).

TheEvilSkeleton:
Isn't it better to have a more restrictive default instead and more aimed to the end user? I personally think that *explicitly* enabling what you want further would be a better idea, since Flatpak lets you do that (and very easily). There are more users than there are developers, which is why I think that focusing on the end user is a better idea. The end user typically only needs `XDG_PICTURES_DIR` for image editing, maybe `XDG_DOWNLOADS_DIR` and very few other directories as well.

> So no we can't and won't remove `--filesystem=host`. This will only happen when we will be able to use the file portal, which we want to do eventually, but will likely not happen immediately (probably not even for GIMP 3.0) because it has limitations which would make us lose features (we need to redesign at the very least the export process to get there, and maybe other features as well).

I see, ~~this is why it's better to read everything before replying to every paragraph.~~

Jehan:
> Isn't it better to have a more restrictive default instead and more aimed to the end user? I personally think that explicitly enabling what you want further would be a better idea, since Flatpak lets you do that (and very easily). There are more users than there are developers, which is why I think that focusing on the end user is a better idea. The end user typically only needs XDG_PICTURES_DIR for image editing, maybe XDG_DOWNLOADS_DIR and very few other directories as well.

Euuh… what I said is exactly about being aimed to the end user 👨‍🎨 . I can assure you that no end users who use GIMP regularly on a wide range of projects want to be restricted to `XDG_PICTURES_DIR` (or `XDG_DOWNLOADS_DIR`). You are only talking about a very particular set of users who have only casual needs and have a few images in their `~/Pictures/` folder, and maybe download some images to edit without ever organizing their files. I can tell you the professional graphist I work with daily organize their files thoroughly in neatly named project folders, sometimes on other data partitions, and so on. All the other graphics professional I worked with (whether they used GIMP or other software) do the same. And they are **not** developers. You can't find more "end user* than these people who actually use GIMP 8 hours a day, every day for past years and years to come.
They have dozens of project folders, some archived in other disks, some ongoing, neatly organized according to their personal organization, and they also have libraries of resources and references (photos, videos, patterns, textures, scans… all of these personally created and customized over the years, then archived for later use, as they need regularly to load some and use them for a project).

So really no, end users don't *only needs* `XDG_PICTURES_DIR` and `XDG_DOWNLOADS_DIR`. They need to be free to load any file from anywhere they wish. The sandboxing situation can only be fixed when we will be able to handle the file portal issues, that's it. Until then, yes there is no real sandboxing in our flatpak, that's definitely true. That comes with the fact that all these sandboxing technologies are pretty new, therefore existing programs need time to adapt (in particular the ones heavily used to edit random files, like GIMP).
This being said, I also remind that this is Free Software and that you can see (and study) the exact code which is built (all build processes on Flathub are fully transparent and available), and even more that this is an official flatpak (from the same team which develops the software). So if you can't trust the file accesses done by this flatpak, you should as well not trust GIMP at all. Not saying that we should not eventually have file sandboxing (again, when we can use the file portal, and only then), but that it's good to put some things into perspective too. 🙂

Jehan:
P.S.: maybe you don't know what we work on, so that's it (an animation film): https://film.zemarmot.net/en/

We also know many other people who work with GIMP as either advanced users or like us professionally. Just go ask them to limit themselves to the `Pictures/` folder and that they should never need to load or save their files elsewhere. 🤣

PR #77, opened 2020-06-15
Make sure the default path for plug-ins and scripts point to the extension point

hfiguiere:
Followup to #75 

This patch make sure that the extension point for plugin and scripts is in the default path for GIMP.

- If either preference hasn't been changed then it will be set to the proper values.
- If one has been changed, it an be reset in the UI and it will have the proper values, but the user will lose the previous one, but that's how the reset work. This case is not very likely though.
- Same with a reset all.

The change to `gimprc.in` isn't necessary at all, but it is for consistency (and in case of upstreaming).
The patch assume GIMP is run in Flatpak. Work needs to be done to have this upstreamable.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -554,6 +554,10 @@
                     "type": "patch",
                     "use-git": true,
                     "path": "patches/0001-desktop-rename-launchable-in-appdata.patch"
+                },
+                {
+                    "type": "patch",
+                    "path": "patches/gimp-extension-path.patch"
                 }
             ],
 	    "post-install": [
--- a/patches/gimp-extension-path.patch
+++ b/patches/gimp-extension-path.patch
@@ -1,0 +1,58 @@
+diff --git a/etc/gimprc.in b/etc/gimprc.in
+index 8442117a24..c4a867018a 100644
+--- a/etc/gimprc.in
++++ b/etc/gimprc.in
+@@ -75,7 +75,7 @@
+ # Sets the plug-in search path.  This is a colon-separated list of folders to
+ # search.
+ # 
+-# (plug-in-path "${gimp_dir}/plug-ins:${gimp_plug_in_dir}/plug-ins")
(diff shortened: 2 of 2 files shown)


hfiguiere:
bot, build

Jehan:
> The change to gimprc.in isn't necessary at all, but it is for consistency

Only because everything is commented out in the system `gimprc`. Remove the leading `#` and that's all this MR needs to do. :-)

>  The patch assume GIMP is run in Flatpak. Work needs to be done to have this upstreamable.

For upstream patch though, a change in code will be needed too, to make it clean. Yet I won't accept the code as is, indeed, because as you say, this assumes Flatpak. We are all-packaging friendly.

I would suggest a configure option (which has to be implemented both for the autotools and meson build system) to add paths to the default system paths. Well it could be either several options (`--with-plug-in-path`, `--with-script-fu-path`, etc.) or a single option (`--with-additional-data-prefix` maybe?) over which various known directory names could be found (`plug-ins/`, `scripts/` and so on).

I prefer the second as it looks more generic and future-proof to me (and works with the flatpak merge feature), but maybe some packaging use cases would prefer completely separated prefixes (but then if it arises, this packager can also make a patch and we'd update)? I don't know, I'd let you choose.

hfiguiere:
> > The change to gimprc.in isn't necessary at all, but it is for consistency
> 
> Only because everything is commented out in the system `gimprc`. Remove the leading `#` and that's all this MR needs to do. :-)

No, sadly. It only work for "reset" for all settings, not the reset for the individual folders. The later just call the functions I patched.

Jehan:
Yeah ok, I think the system gimprc is shadowed by the user gimprc if you ever changed these config values manually (e.g. you changed paths in preferences). That's why it works for full reset, and I guess it would work for a new install too. But a `reset` would indeed likely "remove" this additional path. Anyway just guessing, too lazy to test.

Let's just merge. I'll be expecting some patch upstream in GIMP obviously, as we want to have our GIMP flatpak as vanilla as possible. :-)

PR #78, opened 2020-06-16
Extension point for manual

hfiguiere:
This is the first step of #1 

Unlike I said in my last comment I found a way to build the manual as an extension package this PR only set the extension points for the build.


Jehan:
Thanks!

PR #79, opened 2020-06-17
Set locale-subset to true for the manual

hfiguiere:
Sorry one more PR but I got a clarification on how the locale work and basically setting this attribute should be all that is needed to have the manual languages selectively installed.

Issue #1

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -37,7 +37,8 @@
     "add-extensions": {
         "org.gimp.GIMP.Manual": {
             "version": "2.10",
-            "directory": "share/gimp/2.0/help"
+            "directory": "share/gimp/2.0/help",
+            "locale-subset": true
         },
         "org.gimp.GIMP.Plugin": {
             "version": "2-3.36",


Jehan:
Ok. Description says "If set, then the extensions are partially downloaded by default, based on the currently configured locales. This means that the extension contents should be a set of directories with the language code as name."

It means that if my OS is French, it will download only the French manual by downloading org.gimp.GIMP.Manual, right? Will it be possible to select also a manual specific language? Like can I explicitly download `org.gimp.GIMP.Manual.fr` for instance? Since we don't have as many localization of the manual as we have GUI localizations (81, if I recall, but far less for the manual), we want to allow people to explicitly download a manual in another language (GIMP will also recognize the manual for the currently selected language is unavailable and propose the list of other available manuals).

Yes I know that we have a loot of requirements, GIMP is complex software. :P

hfiguiere:
If it doesn't exist in said language nothing is installed.

Also you can set more than one language in Flatpak (and thus more than one would be downloaded):
  https://github.com/flatpak/flatpak/wiki/Tips-&-Tricks#adding-languages

But I'm not sure how this is integrated at the system level.

A manual install of the package will install all the available languages.

hfiguiere:
Just to clarify, without that flag, the manual will be downloaded in all the languages, and that's ~200MB. With it, only the request languages will be downloaded. If they are missing nothing is. One can manually install the flatpak to get it all though.
(this could be improved, but it's work at the flatpak level)

This is not blocking the manual packaging.

Jehan:
Yeah ok let's just merge. The full manual package with all language is huge (flatpak tells me `876.5 MB`, is that right?), that's why I was initially wary of installing all languages, but in the same time, I want to make it easy to install a given localization even if it's not the one you set for GIMP. So I'm torn.

Anyway let's just roll like this, at least for now, and see how it goes.

hfiguiere:
It say that here too, but it's closer to 200MB

hfiguiere:
I validated that  doing the update for GIMP I got the manual and it only has "en" and "fr" as requested.

PR #81, opened 2020-09-03
Added DRI access

hfiguiere:
GMic complain about GL. So we need enable DRI permissions. I also wonder if GEGL using the GPU wouldn't complain either.


Jehan:
Ok thanks. 👍

Jehan:
For the record, I realized the build failed (I don't get I got a notification for this), so I just triggered a rebuild. The updated build should be available today.

hfiguiere:
this is unfortunate we don't get notifications. It happened to me last week too.

PR #82, opened 2020-09-24
Add libtiff in our build.

Jehan:
The Freedesktop SDK's libtiff is not full-featured so there are TIFF
files our build cannot create or open right now. This has been fixed
upstream but won't be available within the year most likely. So let's
build it ourselves for the time being.
Cf. https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1117

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -516,6 +516,18 @@
             ]
         },
         {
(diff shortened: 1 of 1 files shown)


Jehan:
Build tested, works well. Let's merge!

PR #83, opened 2020-10-03
Wip/jehan/prepare gimp 2 10 22

Jehan:


Jehan:
Hmm… last test build is not over, but I have basically no functional change since the last successful test build. So let's merge and hope for the best!

PR #85, opened 2020-10-05
Testing GNOME 3.38 SDK.

Jehan:


Jehan:
Ok so Webkit builds fail on all platforms. Also GNOME 3.38 SDK does not provide ARM anymore? So sad! 😢

There are still machines on ARM (32-bit) being release nowadays!

hfiguiere:
> Also GNOME 3.38 SDK does not provide ARM anymore? So sad! cry
> 
> There are still machines on ARM (32-bit) being release nowadays!

Freedesktop SDK 20.08 dropped support for it.

Jehan:
> Freedesktop SDK 20.08 dropped support for it.

Yeah I remember some discussion going on about this (in some report, but I can't find it anymore) when I first discovered they removed x86 (which I was already sad about because there are still some old machines perfectly functional, but at least I could understand as there is no new material with this architecture, from what I was told, isn't there? ARM on the other hand, materials are still being released, like the RPI zero if not mistaken).

So even though I did not follow up, this support drop did not surprise me (I knew it was coming some day). Still sad though.

Erick555:
Among reasons behind arm drop (beside its relative unpopularity) was the fact that machine rented for doing ci was expensive and removing arm allow to cut that cost.

Jehan:
> Among reasons behind arm drop (beside its relative unpopularity) was the fact that machine rented for doing ci was expensive and removing arm allow to cut that cost.

Ah I see. That's a practical reason. Still sad, but such a reason definitely makes sense.

What about cross-compilation though? Is it considered not good enough?

Erick555:
AFAIK it was already cross-compiled on 64 bit machine but if it's removed then there are no free machines left to takeover.

hfiguiere:
> What about cross-compilation though? Is it considered not good enough?

TBF cross compilation is tricky in that setup.

The way you can compile for arm on an x86_64 host with flaptak is by using qemu to run everything, and that end up being the easiest albeit with the performance penalty.

Jehan:
> The way you can compile for arm on an x86_64 host with flaptak is by using qemu to run everything, and that end up being the easiest albeit with the performance penalty.

Is it a limitation of Flatpak? Most distributions (at least Debian and Fedora as far as I know) provide ARM/Linux toolchains. You shouldn't need to go through an emulation layer.

Anyway I can definitely understand there are difficulties and some (apparently not so small) costs incurred. These are totally good reasons to me. I hope these can be resolved some day as there are still some ARM 32-bit machines produced. We'll see… 🤷

Issue #91, opened 2020-10-27
GNOME can't diferentiate GIMP stable and beta


fabriciojardim:
STEPS TO REPRODUCE
1. Install GIMP stable from flatpak
2. Install GIMP beta from flatpak
3. Press overview key and search for GIMP

WHAT YOU EXPECT
* Search results should show the 2 versions of GIMP  preferably with some badging identifying the beta icon.
* It should be possible to have both versions of GIMP open side by side.

WHAT HAPPENS
* Search results only shows GIMP beta.
* You can find your stable GIMP if you use the applications dash, but you can't have both open at the same time.

I'm not able to do it myself, but I think this can be fixed in the flatpak manifest.

Thanks!

Jehan:
Yes this is a limitation of the Linux desktop standards hence of flatpak. Basically the specs don't have concept of multiple version of a single application.
See flatpak/flatpak#1109

The solution so far is to switch the visible branch with one of the following command (depending on whether you want to show stable or beta):

```
flatpak make-current --user org.gimp.GIMP stable
flatpak make-current --user org.gimp.GIMP beta
```

You can also run any of the versions from command line even if not the current branch with:

```
flatpak run org.gimp.GIMP//stable
flatpak run org.gimp.GIMP//beta
```

There is a workaround to this issue, which is that we can give to the development version a different ID. Basically the idea is to pretend this is another software altogether. This is ugly and I don't like this, but this is currently the only solution. I will probably do this in a few days.

Jehan:
@alexlarsson Hi! In the end, I'm just trying to follow the "rename the app-id for the beta release" recommendation. I think I did it all properly, with the `rename-*` options as well. Builds work fine.

But then [the build](https://flathub.org/builds/#/builders/51/builds/2285) fails at "Check for AppStream xml" step with this error:

```
stat: cannot stat 'builddir/*/share/app-info/xmls/org.gimp.GIMP.xml.gz': No such file or directory
```

Why does it look for a xml.gz file with the old ID. Does it pull this ID from the repository name (which I can not change obviously)?

When I look at the build log, in the last lines, I clearly see:

```
Processing application org.gimp.GIMP.dev
Saving icon /app/share/app-info/icons/flatpak/64x64/org.gimp.GIMP.dev.png
Saving icon /app/share/app-info/icons/flatpak/128x128/org.gimp.GIMP.dev.png
[...]

PR #92, opened 2020-10-28
Change the application ID for the beta branch.

Jehan:
It's ugly but this is the only way to not have to manually switch the
flatpak branch through boring command lines.
See issue #91 and flatpak/flatpak#1109.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -1,5 +1,5 @@
 {
-    "app-id": "org.gimp.GIMP",
+    "app-id": "org.gimp.GIMP.dev",
     "branch": "beta",
     "runtime": "org.gnome.Platform",
     "runtime-version": "3.36",


Erick555:
This can't work. You have to submit new app to flathub with new id and new git repo.

hfiguiere:
This won't work.

`flatpak run org/gimp.GIMP//beta` to run beta and `flatpak run org/gimp.GIMP//stable` to run stable

Also `flatpak make-current` to select the default one.

Jehan:
> flatpak run org.gimp.GIMP//beta to run beta and flatpak run org.gimp.GIMP//stable to run stable
> 
> Also flatpak make-current to select the default one.

We know. We have already advised people to run these commands ever since this beta branch exists. See for instance the "*Flatpak additional instructions*" in: https://www.gimp.org/downloads/devel/

Even in our release news, we told about these commands: https://www.gimp.org/news/2020/11/06/gimp-2-99-2-released/#beta-flatpak-available

But the simple fact we have to tell about this tells it's sucky workflow because it means it doesn't integrate at all with the desktop. People who want to test the development versions don't want one or the other. They want to use the stable one for their daily work, and be able to run the dev one on a whim. This is possible on Windows (it would probably on macOS too if our packager had packaged the dev version) but not on Linux right now and that's too bad. 😕

Anyway the current proposed trick is to use a different app ID but that will also imply we make a new repo, which is annoying for a few reasons; one of them being that now everyone already registered this beta branch and people currently using it would not get the updates for next versions and would not know why. So I'm holding back this for now.

Maybe later if we add a "notification channel" in GIMP where we are able to send important messages (opt-in). I don't know, we'll see…

org.gimp.GIMP.json
```diff
@@ -1,5 +1,5 @@
 {
-    "app-id": "org.gimp.GIMP",
+    "app-id": "org.gimp.GIMP.dev",
     "branch": "beta",
     "runtime": "org.gnome.Platform",
     "runtime-version": "3.36",
```
farribeiro, line 5: I recommend bump the runtime version

hfiguiere: it's already at 40

farribeiro: Update this development branch, what did i mean, merge master to this branch

PR #99, opened 2021-03-27
Plugins support

hfiguiere:
- Updated the runtime to GNOME 40
- Added DRI access (for anything GL)
- Added plugin support for GIMP 3

I have tested this with GMic that just got support for GIMP 3 (2.99.4)


Jehan:
Awesome thanks!

PR #102, opened 2021-03-29
Build Little-CMS 2.12 ourselves.

Jehan:
This will hopefully fix some crashes we are experiencing in Little-CMS
2.10 currently available in the GNOME 40 runtime.
See: https://gitlab.gnome.org/GNOME/gimp/-/issues/6636

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -310,6 +310,18 @@
             ]
         },
         {
+            "name": "lcms2",
+            "config-opts": [ "--disable-static" ],
+            "cleanup": [ "/bin", "/share" ],
+            "sources": [
+                {
+                    "type": "archive",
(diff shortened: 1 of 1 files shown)


Jehan:
@hfiguiere We have had a spike of crashes (many reports in the last few days) on our flatpak ever since the last update when we bumped from GNOME 3.36 to 40 runtime. No code change in GIMP, only deps change.

See: https://gitlab.gnome.org/GNOME/gimp/-/issues/6636

The traces we got indicate the crash happens in LittleCMS. Either there is a bug in lcms2, or there was one in GIMP but it only got visible with a recent lcms2. Anyway I am hoping on the former and am trying to build the latest lcms2 (2.12 vs. 2.10 in GNOME 40). If it doesn't work, maybe we will have to downgrade instead to 2.9 (like before). Hopefully we won't have to revert the GNOME 40 bump but this is a possibility if we don't find a working combination.

Jehan:
This is confirmed to fix the crash. Cf. discussion from there: https://gitlab.gnome.org/GNOME/gimp/-/issues/6636#note_1069347

One of the crash reporter tested the current published build and this one (only difference between the 2 is Little-CMS being self-built to a higher version). The current published build crashes, not the updated one. So I merged.

PR #103, opened 2021-03-30
Make sure the default path for plug-ins and scripts point to the exte…

Jehan:
…nsion point

(cherry picked from commit 98a4c814a09baa51a622623df89d27b13f0253e2)

@hfiguiere You were saying in another report that you had already packaged G'MIC to the GIMP 2.99 flatpak. So I just tested, and except the fact it doesn't see G'MIC (so I had to add `/app/extensions/plug-ins/GMic/` manually), it just works! That's cool.

With this MR, I cherry-picked your commit (the patch applies cleanly on `master` branch of GIMP), which should take care of the "finding the extension plug-ins" issue. Was there a reason why you didn't propose this earlier? Were you planning to wait a bit before officially announcing flatpak extensions on the beta branch? Even if "just" G'MIC for now, it's still very cool IMO. 🙂

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -515,6 +515,10 @@
                     "commands": [
                         "xsltproc -o desktop/org.gimp.GIMP.appdata.xml.in.in build/flatpak/remove-future-appdata-release.xslt desktop/org.gimp.GIMP.appdata.xml.in.in"
                     ]
+                },
+                {
+                    "type": "patch",
+                    "path": "patches/gimp-extension-path.patch"
                 }
             ],
 	    "post-install": [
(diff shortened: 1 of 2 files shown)


Jehan:
Well it works. Do you want me to hold up on this @hfiguiere ? I don't want to push if you were not adding this on purpose.

hfiguiere:
Yes I changed the patch locally and forgot to mention it.

I extrapolated that `plug-ins` was the directory for plug-ins. While this is the path for Gimp to load plugins, (there are scripts and other) I think I need to repackage the plugin itself.

Looking at the patch, this looks OK to me. But I might need to change the extension point and repackage the plugin. Things would still be in the same location though.

Jehan:
Ok. Feel free to push another version, or merge this one, or anything else which seems right. 👍

hfiguiere:
This one if good for me. It won't be changing. Just the way the plugins will be built. There is currently only one.

Jehan:
> This one if good for me. It won't be changing.

Ok. Let's merge then.

> There is currently only one.

It's already nice as a PoC. Also G'MIC is an important one and also quite a complicated plug-in which makes it good experimentation field of issues with our API, I guess.

PR #107, opened 2021-04-25
Wip/release/2 99 6

Jehan:


org.gimp.GIMP.json
```diff
@@ -417,48 +417,36 @@
...
-                    "type": "git",
-                    "url": "https://gitlab.gnome.org/GNOME/babl.git",
-                    "tag": "BABL_0_1_84",
-                    "commit": "ff6ac9e6a92a16ae5c5e233432c2fcaa22cb1c5e"
+                    "type": "archive",
+                    "url": "https://www.cairographics.org/releases/cairo-1.16.0.tar.xz",
+                    "sha256": "5e7b29b3f113ef870d1e3ecf8adf21f923396401604bda16d44be45e66052331"
+                },
+                {
+                    "type": "file",
```
hfiguiere, line 429: make it "type": "patch" and remove the section below and it should work.

Jehan: Oh really? Flatpak docs doesn't say that type "patch" can have a "url" instead of "path". Let's try. 🙂

hfiguiere: if it doesn't they it should be added to the repository. That's what I would have done in the first place.

hfiguiere: tbf I thought it would work.

Jehan: Well doesn't work:

> (flatpak-builder:3615397): flatpak-builder-WARNING **: 16:33:02.884: Unknown property url for type BuilderSourcePatch

Trying now with "path" but leaving an URL in there.

hfiguiere: As I said, adding the patch to the repository is likely the only way.

Jehan: > if it doesn't they it should be added to the repository. That's what I would have done in the first place.

Yeah that's what I will do if my last test fails again. I wanted to avoid duplication.
Yet this is driving me a bit nuts 🤪 and such a time waste.

Jehan:
Ok done. Thanks @hfiguiere !

Jehan:
@hfiguiere It's very weird. It looks like the release was not published at all. If I look at the flathub builds, why do I see 3 builds: https://flathub.org/builds/#/apps/org.gimp.GIMP~2Fbeta

Looking at start times, the first one is apparently the one started right after the merge, but somehow even though the build is set as "successful", the repo column says "Deleted". Then a second build is said "Published", even though when updating, I don't see a new beta build on Flathub. And for some reason, there is a third build written as "Deleted".

I have no ideas what happened, but it's weird. And bottom line: the actual build doesn't seem to have been published!

Jehan:
Oups, sorry forget what I wrote previously @hfiguiere ! It was published. Somehow I was expecting 2.99.8 but that's 2.99.6 we just released. I'm too much looking at the future, I guess! Anyway I was wrong (the 3 builds are still weird, but the publication did happen). 🙄

PR #108, opened 2021-05-27
Update some dependencies.

Jehan:
Sync with nightly build (from GIMP repository).


hfiguiere:
The new poppler defaults to require boost header for the splash backend.

Jehan:
@hfiguiere What's the splash backend? Is it something we care about in GIMP? If no, maybe we can just skip this option? (not sure if building boost is an ideal solution)

If you feel like it, feel free to push directly to the `wip/release/2-99-8` branch for a fix.

hfiguiere:
it's one of the two backend for poppler. I don't know how you use poppler in Gimp, but `pdftoppm` use it, and without boost header (it just need the headers to be copied) is can be extremely slow. The option has been around for a while and major distro neglected it.

either `-DENABLE_SPLASH=OFF` or `-DENABLE_BOOST=OFF`, the later being the (bad) status quo *if* the splash backend is used.

BTW looking at the code it seems it's all cairo so it should be fine to just disable BOOST.

hfiguiere:
There are some issues on builders at the moment.

Jehan:
@hfiguiere Ok thanks!

Jehan:
bot, build

novomesk:
Hello @Jehan 
I am new to flatpak and having some problems to download necessary components:
```
error: org.gnome.Sdk/x86_64/master not installed
Failed to init: Unable to find sdk org.gnome.Sdk version master
```
I wanted to add libjxl to org.gimp.GIMP-nightly.json file to see whether it will work or not. When you have time, can you try it yourself?
```
        {
            "name": "libjxl",
            "config-opts": [
                "-DJPEGXL_ENABLE_PLUGINS=OFF",
                "-DBUILD_TESTING=OFF",
                "-DJPEGXL_WARNINGS_AS_ERRORS=OFF",
                "-DJPEGXL_ENABLE_SJPEG=OFF",
                "-DJPEGXL_ENABLE_BENCHMARK=OFF",
                "-DJPEGXL_ENABLE_EXAMPLES=OFF",
                "-DJPEGXL_ENABLE_MANPAGES=OFF",
                "-DJPEGXL_ENABLE_SKCMS=ON",
                "-DJPEGXL_ENABLE_JNI=OFF",
                "-DJPEGXL_ENABLE_TCMALLOC=OFF",
                "-DJPEGXL_ENABLE_TOOLS=OFF"
            ],
            "buildsystem": "cmake-ninja",
            "builddir": true,
            "sources": [
                {
                    "type": "git",
                    "url": "https://github.com/libjxl/libjxl.git",
                    "branch": "v0.6.x"
                }
            ]
        },
```

hfiguiere:
the `master` branch of the SDK isn't on flathub.

org.gimp.GIMP.json
```diff
@@ -366,12 +379,37 @@
...
+                "-DJPEGXL_ENABLE_TCMALLOC=OFF",
+                "-DJPEGXL_ENABLE_TOOLS=OFF"
+            ],
+            "buildsystem": "cmake-ninja",
+            "builddir": true,
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/libjxl/libjxl.git",
+                    "branch": "v0.6.x"
```
hfiguiere, line 409: please use a tag and/or a commit, not a branch. Otherwise the build is not reproducible.

hfiguiere (changes requested):

org.gimp.GIMP.json
Jehan: Sure, you are right. I switched to the last commit in this branch (at time of writing), because there seems to be no 0.6.x release yet (hence no tags).

@novomesk Should we really use non-released code? What about building v0.5? Well the beta branch is still a dev release, so it's up to discussion, but maybe for dependencies, it's still better to build released code.

Jehan:
> the master branch of the SDK isn't on flathub.

Indeed, as @hfiguiere says, the `master` branch of the SDK is not on Flathub because it is made for stabler releases (including stabler dev releases) unlike the gnome-nightly flatpak repo. On Flathub, even the beta branch uses a stable runtime (org.gnome.Platform/Sdk 40).

So I'm unsure how you got this error, unless you mixed up with our own "master" GIMP flatpak (the manifest which is in our main repo vs. the manifests in this repo). 🙂

Anyway let me try your rule here.

org.gimp.GIMP.json
novomesk: So replace the branch with

`commit: "20cccbf646158cb128d68db20d12f93d1fb9b6a1"` ?
https://github.com/libjxl/libjxl/commit/20cccbf646158cb128d68db20d12f93d1fb9b6a1 which is now latest in https://github.com/libjxl/libjxl/commits/v0.6.x

Jehan: > So replace the branch with

Yeah I already did the proposed changes.

Can we know how good/bad the changes are between the last release and this dev branch? I.e. is it really worth releasing an unstable dependency (even though it's dev release, or maybe even more! We already have our own unstability to not add others'! 😛)?

novomesk: Latest release of libjxl 0.5.0 is already old and have some already resolved problems.

The intention of the libjxl developers was make the release from 0.6.x branch. I am using the library in my applications and I believe it is quite good snapshot and I really prefer it more than 0.5.0

In the future we will stick to released versions.

novomesk: > What about building v0.5?

I vote against using 0.5 because:
https://github.com/libjxl/libjxl/issues/407 and https://github.com/libjxl/libjxl/issues/405

Jehan: Ok. Let's go with this because it's a dev release and we can also consider JPEG XL as a recent format hence with acceptable irregularities at the start. 🤷

Jehan:
Ok the build with additional JPEG XL support has succeeded. Looking at build logs, it went well, GIMP detected it correctly and the new plug-in got built. I haven't installed/tested the test build yet though.

novomesk:
I tried `flatpak install https://dl.flathub.org/build-repo/59603/org.gimp.GIMP.flatpakref` and GIMP was able to open .jxl files. So it works!

novomesk:
It seems that libjxl 0.6 will be released very soon. There are some new commits in the [v0.6.x branch](https://github.com/libjxl/libjxl/commits/v0.6.x). When it will be released, still use download via git (not .tar.gz sources) because there are submodules too.

Jehan:
MR welcome. ;-)

novomesk:
> MR against `wip/release/2-99-8` branch welcome. ;-)

https://github.com/flathub/org.gimp.GIMP/pull/116

PR #115, opened 2021-09-26
patches: new patch for cairo!245.

Jehan:
This is a patch for a very old issue we had in GIMP about character
cut-off because of inappropriate glyph bounds approximation.
The fix is to remove the approximation altogether.

The commit was slightly different because of another change in the same
piece of code in the main dev branch of Cairo, so this patch is a
version doing the same thing on stable Cairo 1.16.0.

See:
https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/245
https://gitlab.gnome.org/GNOME/gimp/-/issues/325

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -658,6 +658,10 @@
                 {
                     "type": "patch",
                     "path": "patches/cairo-mr114-gimp-issue-6210.patch"
+                },
+                {
+                    "type": "patch",
+                    "path": "patches/cairo-mr245-issue325.patch"
                 }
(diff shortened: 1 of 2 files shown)


Jehan:
Argh webkitgtk is now failing on aarch64?! Nothing changed, still same runtime. What happened? @hfiguiere Any idea maybe?

I see a bunch of redefinition seen as errors, as well as "not declared in this scope" errors, but I really don't feel like diagnosing this horror package. 😱

Jehan:
bot, build

Issue #122, opened 2021-11-22
Bump poppler to 21.11.0 or over


Jehan:
Poppler 21.11.0 was released on November 1st and fixes a bug we have had for 5 years: https://gitlab.gnome.org/GNOME/gimp/-/issues/1007#note_1314785

This report is for us not to forget. If anyone wants to take care of it, please feel free to propose a MR (@hfiguiere @HarryMichal 😉).

Jehan:
Fixed on the beta branch by #123. Will keep this opened until we backport the change to the master branch.

PR #123, opened 2021-11-23
Update dependencies

HarryMichal:
- exiv2 0.27.4 -> 0.27.5
- poppler 21.10.0 -> 21.11.0
- graphviz 2.49.1 -> 2.49.3

Fixes https://github.com/flathub/org.gimp.GIMP/issues/122

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -61,8 +61,8 @@
                     "sources": [
                         {
                             "type": "archive",
-                            "url": "https://exiv2.org/builds/exiv2-0.27.4-Source.tar.gz",
-                            "sha256": "84366dba7c162af9a7603bcd6c16f40fe0e9af294ba2fd2f66ffffb9fbec904e",
+                            "url": "https://exiv2.org/builds/exiv2-0.27.5-Source.tar.gz",
+                            "sha256": "35a58618ab236a901ca4928b0ad8b31007ebdc0386d904409d825024e45ea6e2",
                             "x-checker-data": {
                                 "type": "anitya",
                                 "project-id": 769,
@@ -226,8 +226,8 @@
             "sources": [
                 {
                     "type": "archive",
-                    "url": "https://poppler.freedesktop.org/poppler-21.10.0.tar.xz",
-                    "sha256": "964b5b16290fbec3fae57c2a5bcdea49bb0736bd750c3a3711c47995c9efc394",
+                    "url": "https://poppler.freedesktop.org/poppler-21.11.0.tar.xz",
+                    "sha256": "31b76b5cac0a48612fdd154c02d9eca01fd38fb8eaa77c1196840ecdeb53a584",
                     "x-checker-data": {
                         "type": "anitya",
                         "project-id": 3686,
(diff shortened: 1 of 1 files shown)


Jehan:
@HarryMichal Did you check it works fine? In particular, did you check before/after to confirm it fixes https://gitlab.gnome.org/GNOME/gimp/-/issues/1007? 🙂

Also the same commit on the `master` branch will be appreciated. 😉

HarryMichal:
> @HarryMichal Did you check it works fine? In particular, did you check before/after to confirm it fixes https://gitlab.gnome.org/GNOME/gimp/-/issues/1007? slightly_smiling_face

Only had time to slap the update together. Will test during the week.

> Also the same commit on the `master` branch will be appreciated. wink

Will make it eventually.

Jehan:
Ok. I'll wait for your self-check.

@HarryMichal Since you helped quite a bit here and on the nightly flatpak in the main repo, do you want to help out more maintaining the stable and beta flatpaks? I.e. get push access here.

I'll continue overseeing and looking at changes (and doing some changes myself when needed), but I'm really looking for other people to lift off a bit of the packaging weight off my shoulders. 🙂

Jehan:
bot, build

Jehan:
Tested. It does fix the [PDF antialiasing bug](https://gitlab.gnome.org/GNOME/gimp/-/issues/1007).

Jehan:
Merged. My proposition still holds @HarryMichal if you want to help with maintaining this package on a more regular basis. 😉

HarryMichal:
> Merged. My proposition still holds @HarryMichal if you want to help with maintaining this package on a more regular basis. wink

Thanks @Jehan for merging this! I'd like to contribute more but my month-long silence must have discouraged you from giving me the access. I chose to ignore the ongoing around GIMP and focus on RL. I'll try to get involved a bit more and then we can raise again the topic of the access to the repo.

Jehan:
@HarryMichal I completely understand the need to focus on what's outside GIMP (or computers in general), which is also why it's better to have more contributors (so that everyone can focus a bit more off GIMP thanks to others). So my proposition still holds. 🙂

As far as I know, you are still the most knowledgeable on this whole deps checker thing and you contributed significant updates to both this repo and the main one (where the nightly flatpak is). So you are still a good fit and I am not discouraged if you are still interested. 👍

P.S.: in GIMP, we don't have a requirement of minimum work, only wanting to do good when you do something, being good to others and being open-minded (for healthy technical discussions where we might disagree, but always in good faith and cordial exchanges).

HarryMichal:
> So my proposition still holds.

In that case I'd like to help out :)

Jehan:
Ok cool. I'll create a report to ask to have you added as contributor on this repo.

As already said to @hfiguiere, there is maybe 1 rule we can create about releasing: we should avoid too many successive releases, because it creates too frequent updates which can be a bother for various reasons (each push to `master` or `beta` triggers a new release to the stable or unstable repository). For instance if you plan multiple updates in a row, make them as one MR even if it means delaying it a few more days so that when you push, it makes a single bigger update.

Alternatively, you can push all the MRs in a quick row, then go to https://flathub.org/builds/ and simply cancel the first MR(s), i.e. only keep the last. Let's avoid like 3 updates per week or something. :-)

PR #124, opened 2021-11-29
Update deps

### LINKED ISSUE #122: Bump poppler to 21.11.0 or over

paolostivanin:
* Update GNOME to 41
* Update Poppler to 21.11.0 (fix #122)
* Update poppler-data to 0.4.11
* Update Ghostscript to 9.55.0

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -2,7 +2,7 @@
     "app-id": "org.gimp.GIMP",
     "branch": "stable",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "40",
+    "runtime-version": "41",
     "sdk": "org.gnome.Sdk",
     "command": "gimp-2.10",
     "separate-locales": false,
@@ -377,8 +377,8 @@
                     "sources": [
                         {
                             "type": "archive",
-                            "url": "https://poppler.freedesktop.org/poppler-data-0.4.10.tar.gz",
-                            "sha256": "6e2fcef66ec8c44625f94292ccf8af9f1d918b410d5aa69c274ce67387967b30"
+                            "url": "https://poppler.freedesktop.org/poppler-data-0.4.11.tar.gz",
+                            "sha256": "2cec05cd1bb03af98a8b06a1e22f6e6e1a65b1e2f3816cb3069bb0874825f08c"
                         }
                     ]
                 }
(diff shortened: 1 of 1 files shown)


org.gimp.GIMP.json
```diff
@@ -2,7 +2,7 @@
     "app-id": "org.gimp.GIMP",
     "branch": "stable",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "40",
+    "runtime-version": "41",
```
hfiguiere, line 5: no. don 't. This require plugins to be updated too.

paolostivanin: Ah ok, thanks!

hfiguiere: I was only commenting on the runtime version here. I case that wasn't clear.

paolostivanin: yep I got that :) I saw that there's another update that's updating deps, that's why I closed this. Once the other will be accepted, I'll update the remaining bits!

Jehan:
PR #123 is now merged. Feel free to propose your additional updates @paolostivanin. 🙂

paolostivanin:
hi, sure thing! you mean including also update to gnome 41? or is this still on hold? :)
thanks

Jehan:
@paolostivanin I'm not sure, let's ask @hfiguiere. Maybe make 2 PRs, one for the deps, one for the runtime. Then we can always update the dependencies first and hold, if needed, the runtime (we'll have to update it at some point too anyway).

Jehan:
@paolostivanin Finally just make a single PR with everything, runtime change included and bump the `org.gimp.GIMP.Plugin` version to 2-41. Do you think you could repost it today or this weekend at latest? There is a security reason (though it may not affect us, and no it's nothing related to log4j 😛).

@hfiguiere If you want to know why, please see here (you must be logged in): https://gitlab.gnome.org/GNOME/gimp/-/issues/7615#note_1336808
Unfortunately Github does not even have security flags.

hfiguiere:
I propose to actually stick to `2-40` as a version branch. If we can still update the plugins and rebuild, so that we avoid the problems outlined with update previously where people have to reinstall their plugins since flatpak doesn't handle that case.

Jehan:
> I propose to actually stick to 2-40 as a version branch.

Sure. You are the boss.

@paolostivanin Listen to @hfiguiere recommendation here, not mine. 👍

Jehan:
Ok so we are releasing today 2.10.30 and I'd want to bump the runtime now for a specific dependency. So I am preparing a build with runtime 41 (kept `2-40` for the plug-in version branch though, as discussed): #126

I have not looked in details to the other deps though, because I want minimum problems in my build for now. So new MR for bumping any dependency is still and always welcome. 🤗

Jehan:
So in the end, I don't bump the runtime because it breaks Webkit building!

What I do instead is making sure that we don't use a second-level library with a vulnerability. We are pretty sure we already don't use it anyway, but just to be double-sure, we don't build poppler (the first-level lib which was using it) with it.

Basically: don't bother my last message. I am still welcoming patches bumping to runtime 41 with webkit still building.

PR #125, opened 2021-12-11
Copy the icon on rename: suppress a warning

hfiguiere:
There was a warning gimp couldn't find the gimp icon. Because it was renamed, so this time it gets copied.


Jehan:
Sure. I saw the warning too. Happy to see that's all which was needed to fix this!

hfiguiere:
rename move the icons, sometime the app still want it. :-)

PR #130, opened 2021-12-23
Issue #129 - remove slang from aalib

### LINKED ISSUE #129: Error about slang
When launching GIMP on the command line, since runtime 41, I get this error:

```
/app/lib/gimp/2.0/plug-ins/file-aa/file-aa: error while loading shared libraries: libslang.so.2: cannot open shared object file: No such file or directory
gimp-2.10: LibGimpBase-WARNING: gimp-2.10: gimp_wire_read(): error
```

Not sure what `file-aa` is but the indication is that the plugin won't work.

I filed an issue with GNOME:
https://gitlab.gnome.org/GNOME/gnome-build-meta/-/issues/463

hfiguiere:
Closes #129

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -405,7 +405,10 @@
         },
         {
             "name": "aalib",
-            "config-opts": [ "--disable-static" ],
+            "config-opts": [
+                "--disable-static",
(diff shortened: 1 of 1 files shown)


hfiguiere:
Tested

PR #144, opened 2022-03-30
Update runtime to gnome 42

FakeShemp:


hfiguiere:
no. this require coordination for the plugins, notably GMic.

Jehan:
@FakeShemp I don't think @hfiguiere meant this needed to be closed. But this could not be merged just like this, on a whim (i.e. it "requires coordination"). So it's good but we just won't merge it right now. Maybe later.

Right @hfiguiere ? Let's reopen this and you can merge it when you see fit, i.e. once the plug-ins are ready too?

hfiguiere:
yes we can coordinate

Jehan:
Open let's reopen. Then @hfiguiere just take your time. Nobody will merge this until it's good on plug-in side. 👍

hfiguiere:
bot, build

hfiguiere:
the plugins already work with 42 as it is the same freedesktop-sdk.

So we should be ok to merge whenever.

Jehan:
@hfiguiere Feel free to update when you are happy with the change. :-)

PR #145, opened 2022-05-30
Remove unused vfs dbus talk-name

hfiguiere:
Seee https://docs.flatpak.org/en/latest/sandbox-permissions.html#gvfs-access


HarryMichal:
The explanation in Flatpak docs seems to be clear enough. Merge as you will.

PR #159, opened 2022-10-20
Update runtime to GNOME 43

pcercuei:
It requires bumping BABL and GEGL to their latest versions for the build to succeed.

Signed-off-by: Paul Cercueil <paul@crapouillou.net>

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -2,7 +2,7 @@
     "app-id": "org.gimp.GIMP",
     "branch": "stable",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "42",
+    "runtime-version": "43",
     "sdk": "org.gnome.Sdk",
     "command": "gimp-2.10",
     "separate-locales": false,
@@ -964,8 +964,8 @@
                 {
                     "type": "git",
                     "url": "https://gitlab.gnome.org/GNOME/babl.git",
-                    "tag": "BABL_0_1_92",
-                    "commit": "befaba5cf1509d46fe27ccf609bd576db2bfebdc"
+                    "tag": "BABL_0_1_96",
+                    "commit": "4fcc59632187d595df3443a94c78238af8c4867b"
                 }
             ]
         },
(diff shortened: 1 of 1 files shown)


hfiguiere:
also require to upgrade the plugins. mostly GMIC is the one I expect to break because of Qt and 43 brings a new Freedesktop-sdk

pcercuei:
Ok. I only did a quick test and it seemed to work OK, but I didn't check in-depth.

novomesk:
I will need to upgrade to 43 also in beta branch when we release GIMP 2.99.14

Jehan:
@novomesk You mean for the release or after it?

And for the record, we want to release GIMP 2.99.14 on Sunday. See: https://gitlab.gnome.org/GNOME/gimp/-/issues/8786
@hfiguiere @HarryMichal (next time, I'll add you in the mentions on the release report, sorry it seems I forgot!)

@hfiguiere Do you think the extensions can be verified by then (if they need any change)? In any case, there is no big pressure. It's still a dev release, so we can afford some small breakages or delays in the extensions release.

novomesk:
@Jehan for the release because it is related with the libjxl upgrade to 0.7.0

The 43 has lcms2 2.13, then we can skip building it by ourselves. Lcms2.13 can be used by new libjxl so we can skip building bundles skcms from submodule and save some space.

hfiguiere:
you should remove lcms from the manifest, it's in the 43 SDK

hfiguiere:
alos libtiff 4.1 is built, but 4.4 is in the runtime. Is this specific version needed?

Jehan:
> alos libtiff 4.1 is built, but 4.4 is in the runtime. Is this specific version needed?

Probably not. As I was saying to @novomesk in another report on our tracker, some stuff may date back from times when the runtime didn't have the dep (or the minimum requirement), or possibly because of very bad bugs on specific versions (but in such case, we would likely have added a comment; or we should have).

So yeah, it could be worth testing removing the libtiff build and testing TIFF import and export to see if it works fine.

Also maybe checking if other dependencies are not in a similar case and can be removed. :-)

hfiguiere:
I have GMIC ready in https://github.com/flathub/org.gimp.GIMP.Plugin.GMic/pull/42

It's the more troublesome in general (because of the dependencies)

hfiguiere:
Plugins are go if you want to do so

Jehan:
Ok. Let's merge this one then.

I'm going to make a branch `wip/release/2-99-14` for preparing GIMP 2.99.14 too.

@novomesk @hfiguiere @HarryMichal Feel free to remove whatever is not necessary anymore, or do any other needed things to this branch which we'll merge in 1 or 2 days, much likely.

Jehan:
Ok in the end, as I was anyway in the flatpak repo, I just did a quick sync by diffing `master` and `beta` json files so I made a MR for the `beta` branch in #164 (inclusing removing lcms).

I also remove libtiff from `master` branch in #163.

PR #168, opened 2022-11-13
Upgrade libjxl, enable BMFF in exiv2

novomesk:


novomesk:
Please merge this only after re-bumping GNOME 43

org.gimp.GIMP.json
```diff
@@ -738,6 +741,8 @@
                 {
                     "name": "libde265",
                     "config-opts": [
+                        "--disable-dec265",
+                        "--disable-encoder",
```
Jehan, line 745: So we don't need these because we enable other decoder/encoder which work better and which we prefer over these, is that it?

org.gimp.GIMP.json
```diff
@@ -314,6 +314,9 @@
                 {
                     "name": "exiv2",
                     "buildsystem": "cmake-ninja",
+                    "config-opts": [
+                        "-DEXIV2_ENABLE_BMFF=ON"
```
Jehan, line 318: I'm sure you must have said it somewhere else on some comment, and I can see in your commit on GIMP repo, you said:

> Improves AVIF, HEIF, JPEG XL metadata support.

What does it improve exactly, regarding metadata?

Jehan (approved):
Globally if you tested, it works and it makes things better, I see no problem in this MR. :-)

Just added some questions in review to understand what the various parts are for.

org.gimp.GIMP.json
novomesk: Those are extra tools so we can skip building them. We need just the library.

org.gimp.GIMP.json
novomesk: Without that switch, exiv2 do not support these new formats.
That switch will be default in the future releases anyway, just in the past some people were worried if some patent trolls might claim they invented copying of data from one array to another.

PR #169, opened 2022-11-14
Revert GNOME 43 bump because of glib 2 74 0

Jehan:
@hfiguiere @HarryMichal @novomesk I need to temporarily revert the GNOME 43 bump, at least from the stable flatpak, because the Glib packed in there contains a major crash bug. It used to be a big crash cause which most/all distributions were even patching a month before an actual GLib (2.74.1) got out with the fix.

Anyway we noticed that the GNOME 43 runtime contains the broken GLib so we started to get crash reports already. Therefore, I'm going to revert the runtime update.
I also asked the runtime maintainers to update GLib in there: https://gitlab.gnome.org/GNOME/gnome-build-meta/-/issues/575

@novomesk If I understood, you absolutely want the runtime 43 for the beta flatpak? I mean, it's a dev build, so we can live with the crash, but if we could avoid to get reports for outdated bugs, it would be better.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -929,8 +929,8 @@
                 {
                     "type": "git",
                     "url": "https://gitlab.gnome.org/GNOME/babl.git",
-                    "tag": "BABL_0_1_92",
-                    "commit": "befaba5cf1509d46fe27ccf609bd576db2bfebdc"
+                    "tag": "BABL_0_1_98",
+                    "commit": "d26b50fb2647497b23445b4bfe220cd29b06c535"
                 }
             ]
         },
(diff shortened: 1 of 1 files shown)


novomesk:
@Jehan I just need lcms 2.13+ which would be used by libjxl 0.7.0

It was present in GNOME 43, so we could remove building lcms by ourselves.
I think GNOME 42 has only lcms 2.12

Jehan:
@novomesk Ok. It seems that they will soon do the update in the runtime (cf. answers in my report) so maybe let's hold on for the dev flatpak, which is anyway much less used. I merged this MR for the stable flatpak at least (we already had many report, in just 2 days). We can re-bump the runtime here when GLib is actually updated there.

PR #178, opened 2022-12-24
Keep gimp-console in the flatpak.

Jehan:
Apparently some people want to use GIMP headless from the flatpak. Using gimp-console could be the easiest way then.
See: https://discourse.gnome.org/t/gimp-2-99-how-to-run-in-headless-batch-mode-on-a-server/13097/1

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -831,9 +831,6 @@
                 "-Dbuild-id=org.gimp.GIMP.flatpak.dev",
                 "-Dcheck-update=no"
             ],
-            "cleanup": [
-                "/bin/gimp-console-2.99"
-            ],
             "sources": [
                 {
                     "type": "git",


Jehan:
Note: I hadn't merged this because apparently `gimp-console` doesn't work properly (cf. the forum link above; I haven't tested myself) so I will want to check this first. No need to change  the flatpak if the binary is broken from the start.

brunvonlope:
> I hadn't merged this because apparently gimp-console doesn't work properly 

@Jehan It's not my intention to sound provocative but, since we aren't sure that gimp-console works, could you know why it is actually [**included**](https://gitlab.gnome.org/GNOME/gimp/-/blob/master/build/linux/flatpak/org.gimp.GIMP-nightly.json?ref_type=heads#L920) in the nightly manifest?

brunvonlope:
My previous comment is completely wrong. Ignore it.

PR #184, opened 2023-01-20
Upgrade exiv2, libheif, libjxl

novomesk:


novomesk:
I tried the `flatpak install --user https://dl.flathub.org/build-repo/1011/org.gimp.GIMP.flatpakref` on my computer and it worked.

PR #189, opened 2023-02-02
Update Cairo to version 1.17.8.

Jehan:
The bot fails to build because since this version, Cairo build is meson-only.
See discussion in: https://gitlab.freedesktop.org/cairo/cairo/-/issues/622

Also using a different source URL, which is the one officially listed in the release news: https://cairographics.org/news/cairo-1.17.8/ I hope it won't clash with Anitya database which seems to favor the Gitlab source. But using static download servers is usually better when possible (especially when recommended by the project itself).

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -937,17 +937,17 @@
         },
         {
             "name": "cairo",
-            "buildsystem": "autotools",
+            "buildsystem": "meson",
             "sources": [
                 {
                     "type": "archive",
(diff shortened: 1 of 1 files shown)


Jehan:
For the record, I stopped the build from MR #187 so that people don't get double updates in Flathub. I'll only update our Flathub build after we merge this one.

Jehan:
bot, build

Jehan:
It builds and works fine, but we have an issue with Cairo 1.17.8, which we fixed today: https://gitlab.gnome.org/GNOME/gimp/-/issues/9132

So I'll likely merge this together when releasing GIMP 2.10.34 (i.e. in a few days).

PR #199, opened 2023-02-25
Revert babl for the "white text" bug.

Jehan:
I may have been too quick to publish the Cairo revert. Babl is a likely culprit. Let's try to revert it and take more time to test.


Jehan:
Build in progress: https://buildbot.flathub.org/#/builders/6/builds/25169

PR #225, opened 2023-06-07
Wip/release/2 99 16

Jehan:
Start preparing GIMP 2.99.16 release.


sukahiroaki:
Sorry to intrude here, just a thought (feel free to ignore if you don't agree): Wouldn't it make sense to also bump the runtime version to gnome 44 in this update?

Jehan:
@sukahiroaki Sure, I guess.

@hfiguiere If we bump the runtime, you'd have to re-build the plug-ins, wouldn't you? Anyway stuff have changed in the API and I would not be surprised that the 2.99 plug-ins are broken. So maybe we should not care too much about breaking plug-ins in the dev version?

Jehan:
P.S.: testing bumping the runtime. Let's at least see if there are no obvious blockers when doing so.

hfiguiere:
> @sukahiroaki Sure, I guess.
> 
> @hfiguiere If we bump the runtime, you'd have to re-build the plug-ins, wouldn't you? Anyway stuff have changed in the API and I would not be surprised that the 2.99 plug-ins are broken. So maybe we should not care too much about breaking plug-ins in the dev version?

There are no working plugins for 3 since GMIC has been broken by API change.

novomesk:
I just tried test build with newer commits but got following error at the end:
```
builddir/files/share/appdata/org.gimp.GIMP.appdata.xml: FAILED:
• attribute-invalid     : <release> timestamp is in the future
Validation of files failed
```

novomesk:
I built and installed this manifest locally with `flatpak-builder` but I am still experiencing huge lags described in https://gitlab.gnome.org/GNOME/gimp/-/issues/9446

Jehan:
@novomesk Ah yeah that's one of the problem with Flathub's build sanity check. It verifies the date of the release date (in appstream metadata), which makes sense for actual release build, but not for testing.

If you need a test build, maybe you can just add a commit adding a temporary `"type": "shell"` source in the "gimp" module running a `sed` command line changing the date in the `desktop/org.gimp.GIMP.appdata.xml.in.in` file. Then we can remove or revert the commit before we merge this.

novomesk:
Test build will probably fail because `gegl_minver       = '0.4.45'`. Let's see if there is something else too.

Jehan:
This is the final test build, then I'll merge. If anyone sees things to improve, now is the time.

We won't care about plug-ins for now, as per @hfiguiere's comment. When we'll get closer to RC release with finally a stable API, I'll likely help G'MIC to update their plug-in (though if someone wants to help first, really be my guest!).

hfiguiere:
There is only ony plugin for GIMP 3 and it has been broke for a few releases due to API changes in GIMP. So yeah. Go ahead.

PR #244, opened 2023-09-13
Update runtime to GNOME 44

hfiguiere:


hfiguiere:
This should be merged before 45 is released, because it will automatically EOL 43 and you'll see drove of complaints about an EOL runtime.

I have run it fine. Including plugins.

Jehan:
Code changes visually look fine (just 2 lines!). I was mostly counting on you to test and merge. Since you said you tested and it works fine, then I'm all good for merging.

Maybe we could merge in the same time the few other automatic update MRs by flathubbot though. If anyone has some time to test that they didn't break anything… :-)

hfiguiere:
one thing at a time :-)

hfiguiere:
GNOME 45 being out, I'll look at updating. And beta too. But I expect more issues with 45 as it has a new freedesktop, so this is good for now.

Jehan:
Awesome! :+1:

PR #252, opened 2023-09-21
beta: Update runtime to GNOME 45

### LINKED ISSUE #251: Update beta to GNOME 45
Update beta to GNOME 45

This is currently blocked by this:

https://gitlab.gnome.org/GNOME/gnome-build-meta/-/issues/736

tl;dr GIMP link against libappstream-glib but it is removed in the platform while still in the SDK.

hfiguiere:
- added appstream-glib in because it's missing from the platform

Close #251

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -2,7 +2,7 @@
     "app-id": "org.gimp.GIMP",
     "branch": "beta",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "44",
+    "runtime-version": "45",
     "sdk": "org.gnome.Sdk",
     "command": "gimp-2.99",
     "separate-locales": false,
@@ -49,6 +49,23 @@
         }
     },
     "modules": [
+        {
+            "name": "appstream-glib",
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Dstemmer=false",
+                "-Drpm=false",
+                "-Dfonts=false",
+                "-Ddep11=false"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://people.freedesktop.org/~hughsient/appstream-glib/releases/appstream-glib-0.8.2.tar.xz",
+                    "sha256": "71256500add5048d6f08878904708b3d0c3875f402e0adcd358e91d47dcd8b96"
+                }
+            ]
+        },
         {
             "name": "gexiv2",
             "modules": [
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -735,23 +735,6 @@
             ]
         },
         {
-            "name": "cairo",
-            "buildsystem": "meson",
-            "sources": [
-                {
-                    "type": "archive",
-                    "url": "https://cairographics.org/snapshots/cairo-1.17.8.tar.xz",
-                    "sha256": "5b10c8892d1b58d70d3f0ba5b47863a061262fa56b9dc7944161f8c8b783bc64",
-                    "x-checker-data": {
-                        "type": "anitya",
-                        "project-id": 247,
-                        "stable-only": true,
-                        "url-template": "https://cairographics.org/snapshots/cairo-$version.tar.xz"
-                    }
-                }
-            ]
-        },
-        {
             "name": "qoi",
             "buildsystem": "simple",
             "build-commands": [


hfiguiere:
freedesktop issue it seems.

bot, build

hfiguiere:
the error is in the checker rule. The tarball name is WRONG.

hfiguiere:
removed the f-e-d-c stanza, it's broken.

Jehan:
> removed the f-e-d-c stanza, it's broken.

f-e-d-c? It looks like you mean the "x-checker-data" part (no idea what f-e-d-c means).

Isn't it possible to just fix the URL template or something if it's wrong? This whole automatic check thing is really really practical (despite the annoying duplicate MRs, it still helps us a lot to keep dependencies up-to-date).

hfiguiere:
> > removed the f-e-d-c stanza, it's broken.
> 
> f-e-d-c? It looks like you mean the "x-checker-data" part (no idea what f-e-d-c means).

flatpak-external-data-checker,

 
> Isn't it possible to just fix the URL template or something if it's wrong? This whole automatic check thing is really really practical (despite the annoying duplicate MRs, it still helps us a lot to keep dependencies up-to-date).

appstream-glib is dead anyway.

I have looked at the code and I don't see any easy to fix it. The version output by Anitya is x_y_z. So there is no `${version1}`, etc

Personally I have a much less favourable view of f-e-d-c. Exhibit 1.

hfiguiere:
shall we merge this?

Jehan:
> shall we merge this?

Sure! I'd prefer for the version checker string to be present, but we can always try to fix it afterwards. @HarryMichal can probably help on this!

> Personally I have a much less favourable view of f-e-d-c. Exhibit 1.

Well it may not be perfect, and the duplicate MRs are also small bothers. But so far, it was extremely useful and helpful, so much that the little issues are perfectly acceptable to me. Anyway which software is perfect? 😅

Personally I really appreciate this version checker system.

Anyway feel very free to take some initiative. A runtime update is definitely a good thing, so as long as you confirmed it works fine with the fix and that you tested it, I'm very fine with you merging such fixes.

My ultimate goal is for me to have to do as less as possible in packaging tasks! 😜

So yeah, just merge if it works fine after the update.

hfiguiere:
A runtime update will remove the need to build appstream glib

https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/15691

My take is that GIMP shouldn't depend on this for runtime. At best at build time. Also it's currently unmaintained and about to be deprecated.

As for f-e-d-c, I couldn't find a way to make it work. IMHO the whole thing needs fixing.

Jehan:
> A runtime update will remove the need to build appstream glib

So after runtime 45? Like 46 or something?

In the meantime, don't you want to merge this MR?

> My take is that GIMP shouldn't depend on this for runtime. At best at build time.

It's a runtime dependency. We use the AppStream format as our metadata format for extensions (a new wrapper format to various types of extensions, like plug-ins, brushes, themes and whatnot). This needs to be read at runtime to display list of installed extensions, or installable extensions, etc. (i.e. like installing plug-ins in Firefox, where you have plug-in screenshots, descriptions and more metadata).

> Also it's currently unmaintained and about to be deprecated.

Sure. We need to move on to libappstream. I've heard about this coming deprecation. We'll do this eventually. It's not the first time we change underlying dependencies and it won't be the last. But for the time being, it's what it is and updating this is not my priority for GIMP 3. ;-)

hfiguiere:
> > A runtime update will remove the need to build appstream glib
> 
> So after runtime 45? Like 46 or something?

23.08 update, so 45 too.

> In the meantime, don't you want to merge this MR?

Sure I do.

hfiguiere:
bot, build

hfiguiere:
bot, build

hfiguiere:
Removed Cairo since that version is in the runtime and make f-e-d-c choke because the URL is incorrect.

HarryMichal:
> As for f-e-d-c, I couldn't find a way to make it work. IMHO the whole thing needs fixing.

The tool is certainly quirky and to save myself some energy I just ended up using it using Podman/Docker. Making it run in a manually set up environment would be too much of a hassle.

Also, I'd like express how much I value you (both of you actually) keeping the lights on around here. Life has happened and I am not able to volunteer in the open source world as much as I did and as I'd like to. And also with all that additional cruft around generated by the dependency updater tool.

Jehan:
@HarryMichal No problem. I hope you'll get some more time soon (for FLOSS or just to enjoy it!). :-)

In any case, you are always welcome here to continue contributing whenever you want and if you find the time. Your permissions stay anyway. ;-)

PR #262, opened 2023-11-02
Move HEIC support to an extension

### LINKED ISSUE #111: Move HEIC support to extension
`org.gimp.GIMP` includes HEIF/HEIC support, which links to libx265 and libde265. Some distributors or users may prefer not to have such codecs installed on their system.

Other apps like Totem, Pitivi, Blender, … build a separate `$APP.Codecs` extension, which is automatically installed by default but can be masked by the end-user or distributor to avoid it being installed. They define the extension in their manifest like this:

```json
    "add-extensions": {
        "org.pitivi.Pitivi.Codecs": {
            "directory": "lib/codecs",
            "add-ld-path": "lib",
            "bundle": true,
            "autodelete": true
        }
    }
```

and arrange for (sub)directories of `/app/lib/codecs` to be added to their library and plugin paths, e.g.:

```json
    "finish-args": [
        ...,
        "--env=GST_PLUGIN_PATH=/app/lib/codecs/lib/gstreamer-1.0",
        ...
    ],
    ...
    "build-options": {
        "env": {
            "GST_PLUGIN_SYSTEM_PATH": "/app/lib/gstreamer-1.0/:/app/lib/codecs/lib/gstreamer-1.0/",
            ...
        },
        "append-ld-library-path": "/app/lib/codecs/lib",
        "ldflags": "-L/app/lib/codecs/lib"
    },
```

Then add `post-install` commands to the relevant modules in the Flatpak to move them to subdirectories of `/app/lib/codecs`.

Might it be possible to give GIMP the same treatment? From a brief investigation it seems that `/app/lib/gimp/2.0/plug-ins/file-heif/file-heif` is already a stand-alone module and that other modules do not link to x265/de265 so it may be not too difficult to move it, and the libraries it depends on, to an extension. 

The one downside of this approach is that it is not possible to update the `MimeType` list in the `.desktop` file based on whether the extension is installed. These apps choose to always list the relevant formats. I think this is acceptable – by not installing the extension the user/distributor is deliberately choosing a less-functional app.

wjt:
HEIC is patent-encumbered.  Some distributors or users may prefer not to
have such codecs installed on their system.

libheif has an optional plugin system, where codecs can be compiled as
separate shared objects. At runtime, libheif dynamically loads any such
codecs it finds in its search path. The search path comes out as
/app/lib/libheif, adjacent to the libheif library itself.

Configure libheif to build its libde265 and x265 support as plugins, and
then split the /app/lib/libheif directory into an extension. Move the
libde265 and x265 libraries themselves to a subdirectory, and use
add-ld-path to place /app/lib/libheif/lib onto the library search path
if the extension is installed.

With these changes, the org.gimp.GIMP.HEIC extension will be installed &
uninstalled automatically with the main application, unless the
distributor or user explicitly choses not to install it (via `flatpak
install --no-related`, `flatpak mask`, or some other means). Support for
other formats in a HEIF container, such as AVIF, are always supported;
HEIC is only supported if org.gimp.GIMP.HEIC is installed.
    
Fixes #111.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -39,6 +39,12 @@
         "*.a"
     ],
     "add-extensions": {
+        "org.gimp.GIMP.HEIC": {
+            "directory": "lib/libheif",
+            "add-ld-path": "lib",
+            "bundle": true,
+            "autodelete": true
+        },
         "org.gimp.GIMP.Plugin": {
             "version": "3",
             "directory": "extensions/plug-ins",
@@ -492,15 +498,18 @@
             "config-opts": [
                 "-DWITH_GDK_PIXBUF=OFF",
                 "-DWITH_EXAMPLES=OFF",
-                "-DENABLE_PLUGIN_LOADING=OFF",
-                "-DWITH_LIBDE265=ON",
-                "-DWITH_X265=ON",
+                "-DENABLE_PLUGIN_LOADING=ON",
+                "-DWITH_LIBDE265_PLUGIN=ON",
+                "-DWITH_X265_PLUGIN=ON",
                 "-DWITH_DAV1D=ON",
                 "-DWITH_AOM_DECODER=ON",
                 "-DWITH_AOM_ENCODER=ON",
                 "-DWITH_JPEG_DECODER=ON",
                 "-DWITH_OpenJPEG_DECODER=ON"
             ],
+            "build-options": {
+                "append-pkg-config-path": "/app/lib/libheif/lib/pkgconfig"
+            },
             "cleanup": [
                 "/bin",
                 "/share/thumbnailers"
@@ -510,12 +519,13 @@
                     "name": "libde265",
                     "buildsystem": "cmake-ninja",
                     "config-opts": [
+                        "-DCMAKE_INSTALL_PREFIX=/app/lib/libheif",
                         "-DENABLE_SDL=OFF",
                         "-DENABLE_DECODER=OFF",
                         "-DENABLE_ENCODER=OFF"
                     ],
                     "cleanup": [
-                        "/bin"
+                        "/lib/libheif/bin"
                     ],
                     "sources": [
                         {
@@ -536,13 +546,14 @@
                     "buildsystem": "cmake",
                     "subdir": "source",
                     "config-opts": [
+                        "-DCMAKE_INSTALL_PREFIX=/app/lib/libheif",
                         "-DEXTRA_LIB='libx265-10.a;libx265-12.a'",
-                        "-DEXTRA_LINK_FLAGS=-L.",
+                        "-DEXTRA_LINK_FLAGS=-L/app/lib/libheif/lib",
                         "-DLINKED_10BIT=ON",
                         "-DLINKED_12BIT=ON"
                     ],
                     "cleanup": [
-                        "/bin"
+                        "/lib/libheif/bin"
                     ],
                     "sources": [
                         {
@@ -559,9 +570,9 @@
                         {
                             "type": "shell",
                             "commands": [
-                                "ln -s ${FLATPAK_DEST}/lib/libx265-10.a",
-                                "ln -s ${FLATPAK_DEST}/lib/libx265-12.a",
-                                "rm -fr ${FLATPAK_DEST}/lib/libx265.so*"
+                                "ln -s ${FLATPAK_DEST}/lib/libheif/lib/libx265-10.a",
+                                "ln -s ${FLATPAK_DEST}/lib/libheif/lib/libx265-12.a",
+                                "rm -fr ${FLATPAK_DEST}/lib/libheif/lib/libx265.so*"
                             ]
                         }
                     ],
@@ -571,6 +582,7 @@
                             "buildsystem": "cmake",
                             "subdir": "source",
                             "config-opts": [
+                                "-DCMAKE_INSTALL_PREFIX=/app/lib/libheif",
                                 "-DHIGH_BIT_DEPTH=ON",
                                 "-DEXPORT_C_API=OFF",
                                 "-DENABLE_SHARED=OFF",
@@ -591,7 +603,7 @@
                                 }
                             ],
                             "post-install": [
-                                "mv ${FLATPAK_DEST}/lib/libx265.a ${FLATPAK_DEST}/lib/libx265-10.a"
+                                "mv ${FLATPAK_DEST}/lib/libheif/lib/libx265.a ${FLATPAK_DEST}/lib/libheif/lib/libx265-10.a"
                             ]
                         },
                         {
@@ -599,6 +611,7 @@
                             "buildsystem": "cmake",
                             "subdir": "source",
                             "config-opts": [
+                                "-DCMAKE_INSTALL_PREFIX=/app/lib/libheif",
                                 "-DHIGH_BIT_DEPTH=ON",
                                 "-DEXPORT_C_API=OFF",
                                 "-DENABLE_SHARED=OFF",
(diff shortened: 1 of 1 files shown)


wjt:
I am targeting the `beta` branch because testing a version targeting the `master` branch locally involved compiling WebKit, which I gave up on after half an hour; but I believe the same change would apply there, too.

wjt:
Seems to work! Here's how to test (once the test build is ready):

- Install the test build
- Observe that `org.gimp.GIMP.HEIC` is installed
- Observe that GIMP can still open HEIC and AVIF files
- `flatpak uninstall org.gimp.GIMP.HEIC`
- Observe that GIMP can still open AVIF files, but not HEIC files

Jehan:
This looks very cool. Can you confirm that *by default*, this HEIC extension will be installed when installing the base flatpak? If it needs to be explicitly installed, I can only imagine some frustrated users reporting that the flatpak lost format support. Unfortunately the extension concept is still not so well known by people (the main Flathub page finally shows them, but hidden behind some tab and doesn't even explain how to install them).

Secondly what will happen to people who already had GIMP flatpak installed? When they will update, I guess this extension won't be installed, will it? So these people at least will lose support. Or is there a way so that people who had older builds get this extension auto-installed at their next update only?

wjt:
> This looks very cool. Can you confirm that _by default_, this HEIC extension will be installed when installing the base flatpak?

Correct.

> Secondly what will happen to people who already had GIMP flatpak installed? When they will update, I guess this extension won't be installed, will it?

It will be installed automatically.

wjt:
https://docs.flatpak.org/en/latest/flatpak-command-reference.html#:~:text=no%2Dautodownload%20(boolean,Available%20since%200.6.7. – the `no-autodownload` property defaults to false if not specified, which due to the magic of double-negatives means that autodownload defaults to true.

wjt:
Unfortunately the build is now failing due to a new linter check. The objectionable lines appear to be these finish-args:

```
        "--filesystem=xdg-config/GIMP",
        "--filesystem=xdg-config/gtk-3.0",
```

The linter change is https://github.com/flathub/flatpak-builder-lint/commit/7869b06fd6b70f6cf8d63860274aff9f8925c155

Erick555:
You need to ask for exemption in linter repo: https://github.com/flathub/flatpak-builder-lint/pulls

Jehan:
Done. Thanks @Erick555!

https://github.com/flathub/flatpak-builder-lint/pull/206

hfiguiere:
But why does it need `"--filesystem=xdg-config/GIMP",`?

Jehan:
@hfiguiere It's all explained in https://github.com/flatpak/flatpak/issues/328

Basically we absolutely didn't like that the flatpak had a separate configs, because the biggest users of GIMP are professionals (since it's a pro app) who tweaked their config for dozens of years, have created and downloaded scripts and plug-ins, have custom data (patterns, fonts, brushes, dynamics, shortcuts adapted to their workflows, specific dockable organizations…), and so on. They would use GIMP from flatpak and suddenly it would all be… well not there!

People would have to search the web for why it broke, be unhappy and angry, hopefully understand the problem, then move files by hand.

As I recall, one of the big reason for this separation of config folders too was that if you installed a software from package manager and flatpak, and if they are different versions, the config format may be different. Well we don't really have this problem in GIMP, because we did things well from the start and we versioned our config folders (GIMP 2.8 was in `~/.gimp-2.8/` then 2.10 in `$XDG_CONFIG_HOME/GIMP/2.10/` and GIMP 3 will be in `$XDG_CONFIG_HOME/GIMP/3.0/`). So a 2.10 and a 3.0 version can coexist without a problem without breaking each other's config.

Moreover we have a migration procedure in place (and have had forever). So the first time when GIMP 2.10 is run, it would check if older config folders exist (with specific knowledge to each so if the last config is a 2.6, it can be migrated appropriately, differently to a 2.8, and so on). Without the permission, again, this is broken if their previous run of GIMP was on a non-flatpak.

Now we are in fact considering dropping the permission for GIMP 3. Not because we changed our mind, but because we are not very happy with the current implementation of this special permission. The part we don't like is the *conditional* part: if `$XDG_CONFIG_HOME/GIMP/2.10/` doesn't exist, then the normal flatpak folder is used (in `~/.var/`), otherwise the standard XDG folder is used. I already asked for this to be changed, but this was refused. This is basically random and just makes for headaches. Some people have their config in XDG_CONFIG_HOME, others in .var/, others even in both (when they use a 2.10 both as flatpak and not). That's also a pain to debug when people have issues to ask them to search for their config folders.

Basically absolute logic (always in XDG_CONFIG_HOME, done!), not conditional which creates ugly discrepancies.

For this reason, we might drop the permission for GIMP 3, and we'll just tell people: it's how it is, flatpak sandboxes even your config folders and you can't share them. So you have to copy files manually and keep them in sync yourselves. Because it's just too much a pain otherwise.

But we can't do this in the middle of a series. GIMP 3 will break other things too. This is the right time to also break configs (only for people using flatpak though, fortunately! Others will still have nice config migration and sharing!).

hfiguiere:
bot, build

hfiguiere:
the failure is not caused by the linter.

wjt:
Refreshed this branch to update 2 modules (which is why the build was failing).

@Jehan Any further questions here? As and when this lands on the beta branch, I can prepare a version of these patches for the stable branch.

Jehan:
@wjt Yeah, I just need to take the time to properly test this. These days, I'm spread very thin. I'll try to do it soon(ish). Thanks for the contribution!

wjt:
No problem, just wanted to check you weren't waiting on me!

Jehan:
@HarryMichal @hfiguiere @novomesk If any of you is willing to test this MR, it would help a lot. I think it's a good feature but I really miss time to test thoroughly. Then feel free to merge the MR if it looks like it works as expected. I'm seeing these points:

1. By default, HEIC support is still present.
2. It is possible to uninstall `org.gimp.GIMP.HEIC` extension, then re-running the GIMP flatpak, HEIC support is gone (but not AVIF).
3. Nothing else is broken in either cases 1. or 2.

Jehan:
bot, build

wjt:
I went to rebase this today...

https://github.com/flathub/org.gimp.GIMP/pull/279/commits/62d196d1a1fa54bd58e1b359a89ebeebf3a29d7e updated `libheif` to be built with `-DENABLE_PLUGIN_LOADING=OFF` without explanation. This branch depends on plugin loading being enabled. @novomesk Do you recall why you added this flag?

wjt:
Rebased. My local build is still running, so it may well be that it fails to build here. I'll mark as draft until it's built and tested.

In case it is helpful to ease the merging of this change into the beta branch, if the build succeeds I will split the unrelated commits to a different branch and leave only the HEIC patch here.

novomesk:
> I went to rebase this today...
> 
> [62d196d](https://github.com/flathub/org.gimp.GIMP/commit/62d196d1a1fa54bd58e1b359a89ebeebf3a29d7e) updated `libheif` to be built with `-DENABLE_PLUGIN_LOADING=OFF` without explanation. This branch depends on plugin loading being enabled. @novomesk Do you recall why you added this flag?

It was set to OFF because all decoders/encoders are bundled-in so far.
Feel free to change as needed.

wjt:
I've changed it in this branch. I have verified that it still works as expected:

- By default the `org.gimp.GIMP.HEIC` extension is installed
- With it installed, HEIC image files can be opened
- With the extension manually removed, HEIC image files cannot be opened but AVIF files can still be opened

Jehan:
I think it really makes sense but I'd rather not spend too much time testing this all. @novomesk Would you mind double-checking this? And if it all works as stated, feel free to merge. :-)

novomesk:
Configure dav1d and j2kdec as built-in backend, so they are not built as plugins.

wjt:
Good catch, I made that change.

wjt:
@novomesk any other suggestions for improvements I could make?

novomesk:
bot, build

wjt:
Thank you!

PR #264, opened 2023-11-05
Preparing GIMP 2.10.36 release

Jehan:


Jehan:
Tested. It looks like it's working fine. Merging. GIMP 2.10.36 released!

PR #265, opened 2023-11-08
Exiv2 version 0.28.1 just got released.

Jehan:
For some reason, the build to release GIMP 2.10.36 a few days ago failed (I should have followed through): https://buildbot.flathub.org/#/builders/6/builds/79515

I tried to trigger a new build manually, but now I get new errors because Exiv2 just released a new version: https://buildbot.flathub.org/#/builders/42/builds/7201

So anyway updating Exiv2 to finally be able to upload the new GIMP version.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -327,13 +327,13 @@
                     "sources": [
                         {
                             "type": "archive",
-                            "url": "https://github.com/Exiv2/exiv2/releases/download/v0.28.0/exiv2-0.28.0-Source.tar.gz",
-                            "sha256": "89af3b5ef7277753ef7a7b5374ae017c6b9e304db3b688f1948e73e103491f3d",
+                            "url": "https://github.com/Exiv2/exiv2/archive/refs/tags/v0.28.1.tar.gz",
+                            "sha256": "3078651f995cb6313b1041f07f4dd1bf0e9e4d394d6e2adc6e92ad0b621291fa",
                             "x-checker-data": {
                                 "type": "anitya",
                                 "project-id": 769,
                                 "stable-only": true,
-                                "url-template": "https://github.com/Exiv2/exiv2/releases/download/v$version/exiv2-$version-Source.tar.gz"
+                                "url-template": "https://github.com/Exiv2/exiv2/archive/refs/tags/v$version.tar.gz"
                             }
                         }
                     ]
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -1,6 +1,5 @@
 {
     "app-id": "org.gimp.GIMP",
-    "branch": "stable",
     "runtime": "org.gnome.Platform",
     "runtime-version": "44",
     "sdk": "org.gnome.Sdk",


hfiguiere:
I think the error happened because the tarball might not have been available.

The failure rate of f-e-d-c is high.

Jehan:
> The failure rate of f-e-d-c is high.

Ahah you really have something against this system. Not sure if it's a traumatism or something. 😉

Anyway I don't mind failures to happen. That's just the normal state of things when we deal with stuff as fleeting as internet servers (if you knew how many unexpected failures we have in GNOME CI 😆). What would be nice though would be that Flathub were to send us notifications when an "official" build fails this way.

Anyway the test build failed too, during the "Committing builds" step but I have no idea what the error is and if it's a real error, or against some temporary issue. It says:

> \ Check flathub-hooks has failed
> Job failed: {'check_name': 'flathub-hooks', 'build_id': 62621, 'job_id': 99357, 'status': 3, 'status_reason': 'One or more validations failed.', 'results': '{"diagnostics":[{"refstring":"app/org.gimp.GIMP/aarch64/stable","is_warning":false,"category":"flatpak_builder_lint","data":{"stdout":{"errors":["flat-manager-branch-repo-mismatch"]},"stderr":""}},{"refstring":"app/org.gimp.GIMP/x86_64/stable","is_warning":false,"category":"flatpak_builder_lint","data":{"stdout":{"errors":["flat-manager-branch-repo-mismatch"]},"stderr":""}}]}'}

The only reference I found to this "flat-manager-branch-repo-mismatch" error is the code of `flatpak-builder-lint`: https://github.com/flathub/flatpak-builder-lint/blob/b5f726346e45756d9dcd063034dcd6fcf88143c7/flatpak_builder_lint/checks/flatmanager.py#L67

Is it because the change is in a branch?

Also the whole check was in fact implemented 5 days ago: https://github.com/flathub/flatpak-builder-lint/commit/1544da3a86857e78ee7c8d438624a67c8ea74f2f

hfiguiere:
paging @barthalion 

Only him can explain

hfiguiere:
I think it doesn't like `"branch": "stable",` in the manifest.

hfiguiere:
for the record I usually make it be removed in new package submission reviews.

Jehan:
Ah I also just reported the issue upstream: https://github.com/flathub/flatpak-builder-lint/issues/202

> for the record I usually make it be removed in new package submission reviews.

I don't understand the meaning of the sentence.

hfiguiere:
That is this was a new package submission review I would ask that line 3 be removed.

Now I wasn't aware this had been implemented.

Jehan:
Oh sorry, I had missed the previous comment to the one I cited. With both comments, I understand.

`"branch": "stable"` is a bug? I always thought that was needed. I actually even think this must have come from the original docs/sources I took inspiration from, for the initial manifest, because I see it was in my first commit here.

hfiguiere:
It was never needed and some of the original doc has been forever wrong....

Jehan:
Testing without the branch line. Going to bed. I'll see the result tomorrow.

hfiguiere:
that was it

Jehan:
> We should probably demote it to a warning.

Probably since a lot of people might have the same line as I understand. Or if the line is really unneeded, it would much better to have nicely readable error message. Because `flat-manager-branch-repo-mismatch` is really not clear (initially I thought it was about `git` branches). And even this one error is lost in the middle of what looks like a long json struct. Figuring out the problem is not easy, unless maybe you know Flatpak as well as @hfiguiere, I guess! ;-)

Actually even now, I still don't understand the problem.

According to docs: https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html

> The branch to use when exporting the application. If this is unset the defaults come from the default-branch option.

And "branch" is defined in here: https://docs.flatpak.org/en/latest/using-flatpak.html#identifier-triples

> Flatpak allows architectures and versions to be specified using an object’s identifier triple. This takes the form of name/architecture/branch, such as com.company.App/i386/stable. (Branch is the term used to refer to versions of the same object.) The first part of the triple is the ID, the second part is the architecture, and the third part is the branch.

Well the `master` git branch is what builds the `stable` branch of our package in Flathub. We install GIMP with `flatpak install org.gimp.GIMP//stable`, as opposed to the `beta` branch.

So why was it an error?

Jehan:
So ok, after testing the test flatpak from the link above, it looks like these test flatpak are now to be run with `flatpak run org.gimp.GIMP//test` (whereas it used to be run with branch `//stable` or `//beta` until now). Basically I am guessing that buildbot is overriding the branch name into `test` instead.

But then, I imagine that when the "official" build is made, it uses the expected branch (either "stable" for git `master` or "beta" for `beta` branches). Is that it?

If so, that makes sense. It was always a bit weird that we needed to uninstall the public builds to install the the test builds (and then if you forgot to uninstall the test build, you ended up with a never-updating stable/beta GIMP, which was a bit annoying to explain to our build testers).

Anyway the error is really unclear, as said. I had to double-guess all of this after testing stuff. I don't mind this kind of stuff to be an error rather than just a warning, so that we can improve our flatpak (here, from what I see, the advantages are clearly there; I always thought annoying that the test builds were on the same branch). It's just better with clear errors.

I'd still appreciate a confirmation, though in the meantime, I'll still merge this patch, because this build should have been up for days now and we continue to receive reports for bugs which are fixed in GIMP 2.10.36.

Jehan:
The link to the build for later check that it succeeded to the end, this time: https://buildbot.flathub.org/#/builders/6/builds/80109

Jehan:
For info, I also opened a request asking for notifications on "Official" builds too, not only the test ones: https://github.com/flathub/flathub/issues/4677

PR #282, opened 2024-02-22
Added patch for missing header (from upstream)

hfiguiere:
This is needed to try to build GMIC 3.3.4


hfiguiere:
I needs this to build the plugins. Will merge, the pach can be removed for next release.

PR #306, opened 2024-04-30
Preparing GIMP 2.10.38

Jehan:


hfiguiere:
`patches/gimp-glib-crash.patch` should be removed.

PR #315, opened 2024-06-19
openexr: Sync with nightly manifest

brunvonlope:


brunvonlope:
Tested locally and it's working. Merging

PR #335, opened 2024-08-01
gexiv2, exiv2: Update them

brunvonlope:
… them


hfiguiere:
looks like the new openexr want to download stuff at build time.

brunvonlope:
@hfiguiere Yes. Seems that we will need to add a new module in a separate mr: `libdeflate`

PR #339, opened 2024-08-10
poppler: Update to 24.08

brunvonlope:


brunvonlope:
Tested in both nightly and beta flatpaks and it's working. Merging.

(Just to note, all my previous MRs here were tested in both nightly and beta flatpaks)

PR #343, opened 2024-08-11
graphviz: Update to 12.0.0

brunvonlope:
Also, drop liblab cleanup since it isn't built anymore


brunvonlope:
Tested and it's working fine. Merging.

PR #345, opened 2024-08-12
openexr, imath: Update them

brunvonlope:


brunvonlope:
Tested .exr with different compressions (to see if libdeflate was brokening something) and openexr importing and exporting is working normally. 

Have noticed a metadata warning when loading the one with zip compression but doesn't seem to impact anything so merging.

Issue #334, opened 2024-07-30
Update beta flatpak to Gnome 46


skierpage:
For what it's worth Inkscape is on org.gnome.Platform//46.
Thanks for all you do; onward to 3.0!

brunvonlope:
@skierpage Thanks for reporting. I was planning to update the runtime to 46. But just to note, we don't update the runtime because some other flatpak is doing this, but because certain runtime is the latest version supported, which I don't know for sure.

---

So, I would like to know where we can find official info about the state of support of GNOME runtimes (which have some kind of lifespan as far I know), @Jehan. Do you know?

hfiguiere:
one year from release. 45 will be EOL in October around when 47 is released.

Jehan:
For info, it's already bumped in my GIMP 3.0RC1 MR: https://github.com/flathub/org.gimp.GIMP/pull/287

https://github.com/flathub/org.gimp.GIMP/pull/287/commits/f9ba9c38b95906d57df5deabc7dcacbf32fba050

Jehan:
Oh and to answer your question, I'm not sure where to get detailed info on release versions (the [repo is here](https://gitlab.gnome.org/GNOME/gnome-build-meta) though), but if I recall, I think their maintenance policy is that they support N and N-1 (so up to one version late at a time).

And since GNOME is kinda following Fedora release cycle, i.e. a new version every 6 months, it indeed means support should stop about a year after release.

brunvonlope:
Thank you everyone for the responses. I'm prone to updating things but the runtime is something too big to update right of the bat so it will take time...

... since I first read the changelogs and test the GIMP feature that uses some dep (I like to do this everytime I update a dep). So, let's hope that RC1 will caught me before I can complete this hobby.

brunvonlope:
So, I diffed both 45 and 46 `manifest.json` comparing deps used by GIMP (not diffed the freedesktop runtimes). The main changes are regarding `glib` and `gobject-introspection`, which have changed the way they handle platform-specific GIO API and can impact our goat-exercises, specially the JavaScript one (to be tested).

`libjxl` also changed (its API) so I will test it to ensure that it is working well.

brunvonlope:
Closing as solved

skierpage:
Thanks everyone! After updating to latest org.gimp.GIMP//beta,
```
% flatpak remove --unused 
        ID                                            Branch          Op
[...]

PR #287, opened 2024-03-14
Preparing GIMP 3.0.0 RC1 release

Jehan:
Closes #334

We are not releasing immediately, but I wanted to start testing some permission change before making a decision on whether to drop it completely or use the updated flag.

PR #347, opened 2024-08-18
Update to GNOME platform 46

brunvonlope:
Closes: #334

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -1,7 +1,7 @@
 {
     "app-id": "org.gimp.GIMP",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "45",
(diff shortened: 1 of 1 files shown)


Jehan:
You may also test the version from: https://github.com/flathub/org.gimp.GIMP/pull/287

As said in another comment, I already bumped to GNOME 46 there. :-)

brunvonlope:
@Jehan Thanks for remember me. I will test both

brunvonlope:
Tested flatpaks from this MR and from Jehan one and both have all goat-exercises working and libjxl too (importing and exporting). Just noticed a warning about a broken pipe (is GIMP a organ? 🙂) on libjxl, which is happening to other plug-ins too on master branch, so merging.

PR #321, opened 2024-07-02
Debloat Flatpak

brunvonlope:
This debloats the final flatpak size a bit by not packaging uneeded files.

PR #348, opened 2024-08-19
Don't cleanup gexiv2 headers / pkconfig

hfiguiere:
This is required to build plugins like GMIC

This is a regression from #321

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -128,10 +128,6 @@
                 "-Dtools=false",
                 "-Dvapi=false"
             ],
-            "cleanup": [
-                "/include",
-                "/lib/pkgconfig"
-            ],
             "modules": [
                 {
                     "name": "exiv2",


hfiguiere:
```
Run-time dependency gimp-3.0 found: NO (tried cmake)

../enginePlugin/meson.build:29:11: ERROR: Dependency lookup for gimp-3.0 with method 'pkgconfig' failed: Could not generate cflags for gimp-3.0:
Package gexiv2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gexiv2.pc'
to the PKG_CONFIG_PATH environment variable
Package 'gexiv2', required by 'gimp-3.0', not found
```

hfiguiere:
bare metal:

```
$ PKG_CONFIG_PATH=/app/lib/pkgconfig pkg-config --cflags gimp-3.0
Package gexiv2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gexiv2.pc'
to the PKG_CONFIG_PATH environment variable
Package 'gexiv2', required by 'gimp-3.0', not found
```

brunvonlope:
@hfiguiere This doesn't make sense to me since GIMP isn't a SDK.

Isn't plug-in author responsibility to building the dep?

Jehan:
@brunvonlope GIMP is a SDK, as far as plug-ins are concerned. That's even why on [our website](https://www.gimp.org/), we write:

> GIMP is a high quality framework for scripted image manipulation, with multi-language support such as C, C++, Perl, Python, Scheme, and more! 

GIMP is indeed not only a GUI, it's also a development platform for graphics editing. 🙂

Plug-in authors must build their own additional deps, sure, but if they have any dependency which is already inside our package, it would be a bit of a shame if they had to re-compile it and re-ship it in their own package. In fact, we could even argue if maybe we should not leave all depencies' headers and pkg-config files. Though maybe not (we'll see on case by case basis).

But in any case, we should definitely leave the ones needed by our own API. We are not going to ask plug-in developers to re-compile libgimp (which basically means compiling GIMP) just to compile their plug-in. This commit was one such case (GExiv2Metadata is used in our public API as parent of GimpMetadata, which is why our pkg-config requires GExiv2 too).

brunvonlope:
> GIMP is a SDK, as far as plug-ins are concerned.

> But in any case, we should definitely leave the ones needed by our own API

@Jehan That's clear for babl, gegl and gimp namespaces. I 100% agree with this, which were already preserved in the cleanup MR #321.

---

> (GExiv2Metadata is used in our public API as parent of GimpMetadata, which is why our pkg-config requires GExiv2 too).

@Jehan Thanks for the response, I am satisfied (it is technical so now I understand the rationale, which is really rational).

@hfiguiere The commit message of your MR, however, was so laconic and it was merged so fast (I literally woke up seeing this) that, if it weren't for Jehan's answer, I would have already reverted this, which I will not to because Jehan response made clear that we have a real reason for this. 

---

> Plug-in authors must build their own additional deps, sure, but if they have any dependency which is already inside our package, it would be a bit of a shame if they had to re-compile it and re-ship it in their own package. 

@Jehan Now, if you are talking about shipping headers even from optional deps (not directly used by our API) just to reduce work of maintainers at build time, I think that, since they are not on our namespaces (excluding cases of dependencies used directly by our namespaces, like gexiv2), it's not our responsibility to do this and will not be a mortal sin to say a "no". After all, GIMP flatpak is said to be vanilla, so extensions should not be the de facto rulers of what happens in the GIMP manifest just because the (optional) dependencies coincide (I know that it isn't gexiv2 case).

Again, you already explained that it is not the case of gexiv2. If you are referring to other deps, what can I do? Go ahead and bloat it with some bytes, I will not discuss.

Jehan:
> Now, if you are talking about shipping headers even from optional deps (not directly used by our API) just to reduce work of maintainers at build time, I think that, since they are not on our namespaces (excluding cases of dependencies used directly by our namespaces, like gexiv2), it's not our responsibility to do this and will not be a mortal sin to say a "no".

Well… say a plug-in wants to have access to libjxl API, whose library is shipped in our package. For us adding the header and pkg-config files is just nothing (I mean, it's just text file, that's not what will make our package huge). If we don't do this, plug-ins have to rebuild then re-ship a second libjxl inside their package (we'll have 2 libjxl in the sandbox!).

That's a bit sad IMO. And it's not like it can cause any real issue for us to just leave the headers around. It's not about who is a "ruler" of GIMP manifest. Just avoiding duplication of resources when possible.

But now, that's all theoretical speech for now. Let's just see if anyone ever asks for access to dev files of one of the libs we ship.

> what can I do?

Nothing for now. I was just making some case that it's not unreasonable, even for other deps. But there is no need to do anything until we have a real use case presented to us.

> Go ahead and bloat it with some bytes, I will not discuss.

Honestly, how many "bytes" are we talking about? That's header files we are talking about. Basic small text files, not binary compiled executables.

Anyway as said, that was just for the sake of discussion. I'm not asking to do anything. Let's leave it at it.

hfiguiere:
You can still fix this upstream. All I know is that of 3 plugins 3 fail to build because of that. And I need also #349 for the same reasons.

hfiguiere:
> @hfiguiere The commit message of your MR, however, was so laconic and it was merged so fast (I literally woke up seeing this) that, if it weren't for Jehan's answer, I would have already reverted this, which I will not to because Jehan response made clear that we have a real reason for this.

I needed to rebuild the plugin now that you changed the runtime. And I couldn't. I just woke up to an unecessary bike shedding.

brunvonlope:
>  And I couldn't.

@hfiguiere My GNOME runtime broken the plug-ins? I didn't think of this possibility, sorry.

---

> bike shedding

@hfiguiere Think what you want. From my part, I will still do my MRs in slow pace waiting for your valuable reviews.

Jehan:
> I said I would not discuss but I need to clarify that it isn't true? Indeed the dep will be re-built to get the headers but it can be cleaned like we do, no need to re-ship as far I know.

Of course it needs to reship their own version of the same dep. Rule n°1 in packaging is to avoid running with a different build that the one you built with. Of course, there is the whole API/ABI thing and in theory in many case, it should still work out (i.e. as long as API/ABI stayed stable), but that is fragile and in practice, packagers always rebuild when something in a dep is changed.

And for a case like ours, if a plug-in builds against a different build for a dep which already exists in our package, they should better ship it or be ready for bad surprises at times.

> it's mostly about not creating (in my opinion) questionable precedents.

I said not to do it unless it were requested. I never asked you to create a "precedent". But yes, if we had a request, that would be reasonable.

> My GNOME runtime MR brokened the plug-ins? I didn't think of this possibility, sorry.

Q.E.D. Yes, updating dependencies usually implies a rebuild.

hfiguiere:
Not sure they broke it is still the same fdo-sdk but I'd rather rebuild them.

GMIC is broken for other reasons: the API changed again and they are slow to catch up.

Jehan:
Oh and by the way @hfiguiere:

> You can still fix this upstream.

Nothing to fix. This is normal. `libgimp` relies on GExiv2, just as GExiv2 relies on Exiv2 (for #349). 😃

hfiguiere:
then it's all good. it didn't seem wrong to me. I just didn't catch it.

PR #358, opened 2024-09-15
Enable OpenMP to help suitesparse

brunvonlope:
~~Depends on: https://github.com/flathub/org.freedesktop.Sdk.Extension.llvm18/issues/9~~ not anymore due to 0499290fbc991746408b8f3de4522616e4145fde


brunvonlope:
Foreground select is working normally in my tests, merging.

PR #365, opened 2024-10-27
Update openexr, libdeflate, poppler and suitesparse modules

brunvonlope:


brunvonlope:
Tested in both Beta and nightly manifest and exr and pdf plug-ins and foreground select are working fine. Merging then.

PR #379, opened 2024-12-31
Add 'libthai' module

brunvonlope:
This is needed for 'th' lang word breaking.


brunvonlope:
There is not too much that we can do on GIMP side: https://gitlab.gnome.org/GNOME/gnome-build-meta/-/merge_requests/3334

PR #389, opened 2025-01-26
plug-ins: change the extension path

hfiguiere:
This should help fixing https://github.com/flathub/org.gimp.GIMP.Plugin.Resynthesizer/issues/14

This is an important change and I will update all the GIMP 3 plugins to work.


hfiguiere:
I haven't tested this with RC3, but I don't think it's a problem.

It is important to merge this ASAP to allow merging the plugins. This will fix Resynthesizer.

Issue #369, opened 2024-11-06
Downgrade ibus to 1.5.28


chenxiex:
As described at [https://www.reddit.com/r/linux4noobs/comments/1cbs7ek/comment/l10muc2/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button](https://www.reddit.com/r/linux4noobs/comments/1cbs7ek/comment/l10muc2/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button), the GNOME 46 runtime still has version 1.5.28 of ibus, making ibus 1.5.29 unusable with the error below:
```
(gimp-2.10:2): Gtk-WARNING **: 00:10:03.337: Loading IM context type 'ibus' failed

(gimp-2.10:2): Gtk-WARNING **: 00:10:04.405: /app/lib/gtk-2.0/2.10.0/immodules/im-ibus.so: undefined symbol: ibus_input_context_set_post_process_key_event
```
The related commit is 7ca35e6.

Issue #363, opened 2024-10-01
Can't open iOS 18 generated photos


hadess:
Using the current GIMP Flatpak, and opening files generated by an iOS 18 device, I get either:
`Loading HEIF image failed: Invalid input: Unspecified: Metadata not correctly assigned to image`
or
`Loading HEIF image failed: Invalid input: Unspecified: Too many auxiliary image references`
as error messages.

iOS 18 support is one of the things that the 1.18 branch of libheif adds:
https://github.com/strukturag/libheif/releases/tag/v1.18.0

735387ade8faa686a07863398494ac8100af0d3e says libheif is "abandoned" forcing < 1.18.0 versions, but doesn't link to why newer versions aren't considered (either here, or in the GNOME GitLab).

brunvonlope:
I suppose that libheif module stopped to be updated in the stable flatpak due to API incompatibilities. 

We use the most recent libheif on the beta one.

hadess:
> libheif module stopped to be updated in the stable flatpak probably due to API incompatibilities. See: [#338 (comment)](https://github.com/flathub/org.gimp.GIMP/pull/338#issuecomment-2285140265)

I couldn't find an upstream bug report, or anything else that would explain what this "incompatibility" was. Do you have a reference for it?

brunvonlope:
> I couldn't find an upstream bug report

Long response: in the linked MR I observed that building libheif with the version mentioned in the MR description was making GIMP unable to export to HEIC and AVIF. First, I just ignored the bug but them flathub bot was becoming annoying so in a further MR I "abandoned" the module to avoid flathub creating new automatic MRs.

I don't know the reason for it so I can't explain nothing. If I know, I would pushed a fix. The "API" thing mentioned in this issue is merely a guess. And I didn't opened any issue about this.

hadess:
The beta worked for my needs. I would say that we should leave this open until beta is the stable branch, so it registers as a known issue, and we can work on backporting whatever fixes are necessary if somebody wants it fixed properly.

Issue #361, opened 2024-09-28
Update runtime to GNOME 47


x80486:

brunvonlope:
I would like to wait at least Ubuntu 24.10, which is a base to some rolling release distros, to be released. Right now, updating the runtime too fast can be tiresome to our downstream maintainer.

PR #399, opened 2025-02-27
GIMP 3.0 Release

### LINKED ISSUE #71: Remove unsafe python2 and webkitgtk1
We should seriously consider removing these unsafe and insecure dependencies. This is even more critical when considering that the GIMP flatpak isn't sandboxed, since it uses `--filesystem=host`.

WebKitGTK 1 is used in the help browser, but has not been supported since 2016, and has dozens of CVEs against it. Without WebKitGTK 1, opening GIMP help will just open up a link to the help pages with the default browser, which in my IMAO is a fine trade-off in the name of security.

Most every single distro has already removed WebKitGTK 1 as a GIMP dependency, i.e. see [debian stable](https://tracker.debian.org/media/packages/g/gimp/control-2.10.8-2).

Python 2 is EOL as of January 1, 2020, and is no longer receiving security updates. Obviously [Arch](https://www.archlinux.org/packages/extra/x86_64/gimp/) has already removed python2 as a GIMP dependency, but even conservative distros are already removing python2. Both the [debian unstable](https://tracker.debian.org/media/packages/g/gimp/control-2.10.18-1) and [debian testing](https://tracker.debian.org/media/packages/g/gimp/control-2.10.14-3) GIMP packages no longer use python2 as a dependency.

I would be happy to submit a pull request to remove both of these dependencies, but I wanted to get some feedback and thoughts first.

brunvonlope:
Closes #71 #361 #363 #369

---

@Jehan This time, I will make the flatpak release


Jehan:
By the way, maybe we should also update the `beta` branch when we release to `master`. Basically both `beta` and `master` would be the stable release (until the next dev release). Otherwise people using `beta` branch would be stuck for months on a build older than stable.

What do you think, @brunvonlope ?

brunvonlope:
> What do you think, @brunvonlope ?

> it's probably wise to release it on beta first (or simulatneously)

Agree. One first step is #400 and other MRs that I will merge

Jehan:
Thanks @hfiguiere. By the way, not sure if you follow on our Gitlab, by the "hope" is to release on Sunday (so far, looks like it will work out, as we haven't had any huge blocker which we haven't been able to quickly take care of so far).

hfiguiere:
there is no runtime 48 and please don't do this without coordinating as the plugins need to follow.

brunvonlope:
> there is no runtime 48 and please don't do this without coordinating as the plugins need to follow.

@hfiguiere Man, nothing was merged yet. There are 5 days remaining.

brunvonlope:
@hfiguiere If the build goes alright, we will use GNOME Runtime 48 on GIMP release today. If you don't have time to update your plug-ins we can help you. Thanks in advance.

Erick555:
You will have heic support in ffmpeg-full extension. Actually gimp was [mentioned](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1704) as one of the targets of this functionality.

brunvonlope:
> gimp was [mentioned](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1704) as one of the targets of this functionality by person who implemented current heic support in gimp that you linked

Happy to know, could you create an issue to we not loose track of it?

Erick555:
Done here: https://github.com/flathub/org.gimp.GIMP/issues/414

PR #406, opened 2025-03-15
README: Update "How to do a flatpak release" following our practice

brunvonlope:
See commit message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,34 +24,20 @@ $ git checkout -b wip/release/2-10-24
 * In the manifest `org.gimp.GIMP.json`, update:
 
   - the "runtime-version" if needed.
-  - the version of "org.gimp.GIMP.Manual" extension with the
-    `major.minor` version of GIMP.
-  - the "version" of "org.gimp.GIMP.Plugin" extension with the following
-    scheme: GIMP API (2 or 3) and the Runtime versions (40 for GNOME 40
-    runtime for instance). 
-  - any dependency module which might be outdated.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,34 +24,20 @@ $ git checkout -b wip/release/2-10-24
 * In the manifest `org.gimp.GIMP.json`, update:
 
   - the "runtime-version" if needed.
-  - the version of "org.gimp.GIMP.Manual" extension with the
-    `major.minor` version of GIMP.
-  - the "version" of "org.gimp.GIMP.Plugin" extension with the following
-    scheme: GIMP API (2 or 3) and the Runtime versions (40 for GNOME 40
-    runtime for instance).
```
Jehan, line 31: The "org.gimp.GIMP.Plugin" version scheme changed apparently. It apparently used to be major + the runtime version. Now it's only major version.

Is that normal?

@brunvonlope @hfiguiere 

@brunvonlope

brunvonlope: @Jehan Since 34b2f8c060cde4c8eefe3ef520cc147763d92938 it is just GIMP major version. From what Hubert said some times, seems that just rebuilding the plug-ins with the new runtime is enough.

hfiguiere: there is not point having more than that. It's too much pain for maintainers and users.

The only plugin that have difficulties with runtime changes is the one that builds Qt5. (GMIC)

Jehan: Ok. Thanks @brunvonlope and @hfiguiere !

PR #416, opened 2025-03-19
Try to downgrade pixman.

Jehan:
Our flatpak has 0.44.2 but I can't reproduce with 0.43.4 on my system and valgrind shows weird invalid writes of pixman.

It's probably not this though, because pippin also has 0.43.4 and can reproduce. But it's worth testing anyway.


Jehan:
Was not this. Closing.

PR #437, opened 2025-06-05
Enable dependency updates for both branches

martymichal:
So that we don't have to sync the two branches manually. @brunvonlope, @hfiguiere, thoughts?

### DIFF
--- a/.github/workflows/update.yaml
+++ b/.github/workflows/update.yaml
@@ -1,0 +1,26 @@
+name: Check for updates
+on:
+  schedule:
+    - cron: '0 7 * * *' # Run once a day, on Monday, at 07:00
+  workflow_dispatch: {}
+jobs:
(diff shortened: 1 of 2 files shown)


brunvonlope:
@martymichal That's a nice idea but have only one problem: by freely allowing the robot to do automatic updates we assume the risk of brokening the flatpak or introducing bugs too often. Some dependencies like suitesparse and libx265 I are pretty problematic.

My actual workflow is the following: 1) I check the changelog of the new version to see if there is some build option to be updated; 2) test the affected gimp feature (for example: in poppler case pdf importing and exporting) to see if there is no regression. 

PS.: Right now, I am deliberately updating the stable branch too frequently just to wipe GIMP 2.10 commits from the remofe OSTree repo. But in the future I will hold up a bit like before for the reasons above.

PR #438, opened 2025-06-05
poppler: Update poppler-25.05.0.tar.xz to 25.06.0

martymichal:


martymichal:
The build seems to work alright. PDF loading and exporting works.

Issue #414, opened 2025-03-17
Replace libheif and its deps in favor of `org.freedesktop.Platform.codecs-extra` on GNOME 49


Erick555:
Since gnome 47 and later the `org.freedesktop.Platform.ffmpeg-full` extension includes HEIC support. It can replace internal `org.gimp.GIMP.HEIC` extension and bundling libheif and its dependencies in gimp.

https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1704

https://github.com/flathub/org.gimp.GIMP/pull/262

Erick555:
Instructions about using the extension can be found here: https://docs.flatpak.org/en/latest/extension.html#loading-existing-extensions

brunvonlope:
Looking better at the docs, seems that I need to point to the "extension" version. This would be even harder to maintain than the Clang sdk-extension, which we dropped (https://github.com/flathub/org.gimp.GIMP/commit/0499290fbc991746408b8f3de4522616e4145fde) because it was a absolute hell to match GNOME Runtime with Freedesktop SDK, specially on the Nightly repo

Erick555:
No, it should be easy. Version should match Freedesktop SDK version that gnome uses. For gnome 48 version will always be 24.08. For gnome 49 and beyond the extension will be handled automatically by runtime and it could be entirely removed from app manifest. So as long as you use gnome 48 use 24.08 extension and on next runtime update you remove it completely.

Below should work right now:
```

    "add-extensions": {
        "org.freedesktop.Platform.ffmpeg-full": {
            "directory": "lib/ffmpeg",
            "version": "24.08",
            "add-ld-path": ".",
            "no-autodownload": false
        }
    },
    "cleanup-commands": [
        "mkdir -p /app/lib/ffmpeg"
    ],
```

Erick555:
The only obstacle I see is that gimp is currently doing some [multi-stage build](https://github.com/flathub/org.gimp.GIMP/pull/88) for x265. This is something runtime is [lacking](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/elements/extensions/codecs-extra/libx265.bst). Perhaps it could be replicated there if someone sends MR.

brunvonlope:
Don't take me wrong, I would do love to not have to build x265 anymore (it's a problematic dep) and libheif too. 

[...]

PR #509, opened 2025-09-16
manifest: Update runtime-version to GNOME 49

brunvonlope:
Closes: #414

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -1,7 +1,7 @@
 {
     "app-id": "org.gimp.GIMP",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "48",
+    "runtime-version": "49",
     "sdk": "org.gnome.Sdk",
     "command": "gimp",
     "build-options": {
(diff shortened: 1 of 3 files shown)


brunvonlope:
bot, build

brunvonlope:
Pipeline succeeded. This will be merged soon

hfiguiere:
How about wait that I say this is ready? 

I clearly wrote 

> But I need a heads up for GMIC that needs to be ready to merge

But I never said it was. If you want me to stick to a schedule, I can invoice.

Alternative. I can just stop caring.

brunvonlope:
> How about wait that I say this is ready?

I think this is as vague as the heads up I did 3 days ago 😅

One week due date, maybe? From my experience maintaing this flatpak (which is way complex than GMIC plugin), this is more than needed to fix problems with dependencies etc.

Jehan:
@hfiguiere I was just suggesting some idea, if this doesn't for with you, just tell us. No need to bite us. You know we are all mostly volunteers around here, right? 🤗

We just want to find some solution which suits everyone.

Anyway, I guess we can wait a bit longer the next time @brunvonlope ?

brunvonlope:
> Anyway, I guess we can wait a bit longer the next time @brunvonlope ?

Yes, certainly. If we don't agree with any "due time", next time I will use the 1 week as a midterm to merge. IMO we should also consider that the runtime release date is well know since the schedule is public: https://release.gnome.org/calendar/

Is just this specifc time the new 49 runtime fixed some long standing issues with HEIC support so I was a bit more fast to merge.

PR #523, opened 2025-09-28
Change filesystem option from host to xdg-pictures

eternaltyro:
Remove write access to `host` and replace with write access to `xdg-pictures` . This follows good practice to make the sandboxing more robust. If `xdg-pictures` is not the right path or an adequate path, we can add an additional one. 

When I try to save or export a project, only the allow-listed filesystem path are shown in the file browser.

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -22,7 +22,7 @@
         "--socket=fallback-x11",
         "--socket=wayland",
         "--device=dri",
-        "--filesystem=host",
+        "--filesystem=xdg-pictures",
         "--filesystem=xdg-config/GIMP:create",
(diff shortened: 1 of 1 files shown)


eternaltyro:
Flatpak is still evolving and there are limitations in how it handles sand-boxing. I think we should still consider this change.

Jehan:
This is indeed not the correct way to handle the sandboxing issue. GIMP is a pro tool and people need to be able to load files from anywhere. Many people have mount points out of their home folder.

The right fix for this issue will be when we will finally implement the native file dialog portal in GIMP. This will allow us to go across a very tight file system sandboxing in a much more secure way. Tightening the sandbox only through permissions like this will only frustrate advanced users without even bringing a lot more security.

PR #722, opened 2026-08-08
Make gtk-config dir as read-only

Erick555:
Write access shouldn't be needed there, see https://github.com/flathub/org.gtk.Gtk3theme.Breeze#workarounds

### DIFF
--- a/org.gimp.GIMP.json
+++ b/org.gimp.GIMP.json
@@ -25,7 +25,7 @@
         "--device=all",
         "--filesystem=host",
         "--filesystem=xdg-config/GIMP:create",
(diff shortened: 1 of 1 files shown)


brunvonlope:
Hi. Thank you for the patch. It is right.

Could you make a PR on https://github.com/flathub-infra/flatpak-builder-lint/blob/master/flatpak_builder_lint/staticfiles/exceptions.json too?

Erick555:
done in https://github.com/flathub-infra/flatpak-builder-lint/pull/1104

brunvonlope:
bot, retry

brunvonlope:
bot, build

PR #740, opened 2026-09-01
Release GIMP 3.2.6

brunvonlope:


brunvonlope:
bot, build

Jehan:
bot, build

Jehan:
bot, ping admins

Jehan:
Dear admins!

I don't understand the flatpak error. It says:

> + flatpak-builder-lint --gha-format --exceptions --exceptions-repo stable repo repo
> Error: Appstream: 'E:org.gimp.GIMP.metainfo.xml:releases-not-in-order The releases are not sorted in a latest to oldest version order. This is required as some tools will assume that the latest version is always at the top. Sorting releases also increases overall readability of the metainfo file.'

I wondered if that were a new rule so I reordered all `<release/>` tags by date (they were already ordered, but by release version instead of dates; which personally I preferred as we can group relative stable versions and dev versions; but whatever!) in [this commit](https://gitlab.gnome.org/GNOME/gimp/-/commit/d0c18031e22cc390852b90b7ca427be1fd6f04a1) which is what the current metadata is pointing to for testing (it will point to a tag for the actual release).

But apparently there is something else… The error still mention a problem with release sorting.

Jehan:
Ok in the end, I found the issue. It was in the automatic adding of a `<release/`> tag as "snapshot" tag when the meson version does not match the latest AppStream version. Unfortunately if the latest appstream was a preparation of the upcoming version, it was a problem, regarding version order. 😅

I'm letting the full test build to finish, but I see the Aarch64 one succeeded already (while it was always failing before) so it should be good to go. :-)

Jehan:
Uh. Why was the build suddenly cancelled?! Grmmbl… What a waste…

Jehan:
bot, build

brunvonlope:
Build 319151 worked fine (https://gitlab.gnome.org/GNOME/gimp/-/work_items/16712#note_2867211) and  I tested 319292 which is working fine too. Merging.

